PR #28, opened 2022-10-24
created basic graph class and implemented global mincut algorithm

barkincavdaroglu:
Implemented the probabilistic global min-cut algorithm as described [here](https://dl.acm.org/doi/pdf/10.1145/234533.234534).
Also implemented a graph data structure to represent the network data.


ajmcquilkin (changes requested):
This will be re-implemented in Rust -- DO NOT MERGE

PR #45, opened 2022-10-25
Implemented Graph DS in Rust

barkincavdaroglu:
First part of converting modules to Rust from JS. This PR contains an implementation of a graph data structure that will be used by algorithms. For more info, please view the ticket that is linked to this PR.


src-tauri/src/graph.rs
```diff
@@ -0,0 +1,317 @@
+use std::collections::{HashMap, HashSet};
+
+#[derive(Debug)]
```
ajmcquilkin, line 3: Is this something we need to keep in the codebase?

src-tauri/src/main.rs
```diff
@@ -32,6 +34,7 @@ fn greet(name: &str) -> String {
 }
 
 fn main() {
+    // let mut g = Graph::new();
```
ajmcquilkin, line 37: Nit: remove commented line

ajmcquilkin (approved):
Looks good! Just a few minor changes for code clarity

src-tauri/src/graph.rs
barkincavdaroglu: We need this to see the output of print statements in tests. I think it's best to keep them for now and I can remove later.

ajmcquilkin (approved):

PR #46, opened 2022-10-25
Update README.md

ajmcquilkin:
Updated the README to give more detail on the project and explain the application development flow.

image
*New README look*

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,23 @@
-# Meshtastic-emergency-response-client
+# Meshtastic Emergency Response Client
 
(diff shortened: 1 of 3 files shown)


barkincavdaroglu:
Just checking: do we need a build command for the rust files if/when we add a new dependency?

barkincavdaroglu (approved):

ajmcquilkin:
In what sense? We'll be adding dependencies with `cargo install`, but once one person PRs that change in, all required packages will be installed automatically for everyone when calling either of the `rust:*` commands

PR #49, opened 2022-10-28
Graph DS using Petgraph Library

barkincavdaroglu:
The implementation in https://github.com/ajmcquilkin/Meshtastic-emergency-response-client/pull/45 was not working properly with algorithms. We decided to use [StableGraph](https://docs.rs/petgraph/latest/petgraph/stable_graph/struct.StableGraph.html) from Petgraph lib which is adjacency list representation of a graph.


ajmcquilkin (approved):
Looks good!

PR #53, opened 2022-10-28
Implemented Tarjan's Algo for Articulation Points

barkincavdaroglu:
This algorithm will find articulation points in the graph which are nodes whose removal disconnect the graph. More info about the algorithm and the problem can be found [here](https://www.geeksforgeeks.org/articulation-points-or-cut-vertices-in-a-graph/).

### DIFF
--- a/src-tauri/src/articulation_point.rs
+++ b/src-tauri/src/articulation_point.rs
@@ -1,10 +1,6 @@
-use defaultdict::DefaultHashMap;
-use petgraph::visit::IntoNeighbors;
-
-use crate::graph_p::Edge;
 use crate::graph_p::Graph;
-use crate::graph_p::Node;
 use core::cmp::min;
+use defaultdict::DefaultHashMap;
 
 pub fn articulation_point_helper(
     graph: &Graph,
@@ -59,7 +55,6 @@ pub fn articulation_point_helper(
 }
 
 pub fn articulation_point(graph: Graph) -> Vec<petgraph::graph::NodeIndex> {
-    // hashmap
     let mut disc = DefaultHashMap::<String, i32>::new();
     let mut low = DefaultHashMap::<String, i32>::new();
     let mut visited = DefaultHashMap::<String, bool>::new();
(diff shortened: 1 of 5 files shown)


src-tauri/src/articulation_point.rs
```diff
@@ -0,0 +1,164 @@
...
+        G_2.add_edge(v4.clone(), v5.clone(), 1.0);
+
+        let articulation_points = articulation_point(G_2.clone());
+        let len_articulation_points = articulation_points.len();
+        for node in articulation_points {
+            let node = G_2.g.node_weight(node).unwrap();
+            println!("Articulation Point: {}", node.name);
+        }
+
+        assert_eq!(len_articulation_points, 2);
```
ajmcquilkin, line 162: Nit: might be worth testing the `name` property of the articulation points in addition to the length of the vector

src-tauri/src/articulation_point.rs
```diff
@@ -0,0 +1,164 @@
...
+    return articulation_points;
+}
+
+// Create a unit test for the Graph struct
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn initialize_graph() {
```
ajmcquilkin, line 100: Instead of having `G1` and `G2`, I would probably recommend breaking this test up into multiple tests. Also (nit), I would give the test a more descriptive name

src-tauri/src/graph_p.rs
```diff
@@ -314,6 +331,26 @@ impl Graph {
         edges
     }
 
+    pub fn get_neighbors(&self, node: String) -> Vec<Node> {
+        let node_weight = self.node_idx_map.get(&node).unwrap();
+        let mut neighbors = Vec::new();
+        for neighbor in self.g.neighbors(node_weight.clone()) {
+            neighbors.push(self.g.node_weight(neighbor).unwrap().clone());
```
ajmcquilkin, line 338: What does this `node_weight` function do? It might be worth adding some quick docs on each graph function if it's not too much of a hassle ([reference](https://doc.rust-lang.org/stable/rust-by-example/meta/doc.html)).

ajmcquilkin (approved):
Looks good! Just some minor documentation and style feedback

ajmcquilkin (approved):

PR #59, opened 2022-11-02
Added initial map

sherry1iu:
image

Added basic map to the tauri

Hope to implement additional map styles, maybe something like below:
https://docs.mapbox.com/api/maps/styles/#mapbox-styles


ajmcquilkin (approved):
Nice work!

PR #62, opened 2022-11-02
Implemented Karger's Algorithm for finding global mincut

barkincavdaroglu:
This probabilistic algorithm finds the global mincut of a graph by contracting an edge until the order of graph is smaller than or equal to 2. The sum of weights of edges between these two nodes is the weight of global mincut. 

Note: The algorithm only returns the weight of mincut. I will need to find a way to recover which original vertex belongs to which cut.

### DIFF
--- a/src-tauri/src/globalmincut.rs
+++ b/src-tauri/src/globalmincut.rs
@@ -52,67 +52,6 @@ pub fn random_select(G: &Graph) -> Edge {
     return edges[i].clone();
 }
 
-/*
-pub fn contract_edge(G: &mut Graph, edge: &Edge) {
-    {
-        let u_idx = edge.u.clone();
-        let v_idx = edge.v.clone();
-
-        let node_u = G.g.node_weight(u_idx).unwrap();
-        let node_v = G.g.node_weight(v_idx).unwrap();
-
-        let u_weighted_degree = node_u.optimal_weighted_degree;
-        let v_weighted_degree = node_v.optimal_weighted_degree;
-
-        let edge_weight = edge.weight;
-
-        let new_value = u_weighted_degree + v_weighted_degree - 2.0 * edge_weight;
-
-        G.change_node_opt_weight(u_idx, new_value);
-        G.change_node_opt_weight(v_idx, 0 as f64);
-    }
-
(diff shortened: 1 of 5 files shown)


src-tauri/src/globalmincut.rs
```diff
@@ -0,0 +1,163 @@
+#![allow(dead_code)]
```
ajmcquilkin, line 1: Why do we need this directive?

src-tauri/src/globalmincut.rs
```diff
@@ -0,0 +1,163 @@
...
+    }
+}
+
+// Create a unit test for the Graph struct
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn run_globalmincut() {
```
ajmcquilkin, line 109: What are we looking for in this test case? I don't see that any conditions are being asserted

ajmcquilkin: To be fair, I'm not really sure what we can assert with a probabilistic algo other than that it returns a valid node

ajmcquilkin (approved):
Looks solid! Just a few questions on testing and on tags

src-tauri/src/globalmincut.rs
barkincavdaroglu: Yeah we cannot test any of that right now. The test is to just print out the success rate.

src-tauri/src/globalmincut.rs
barkincavdaroglu: I was getting warnings (function x is never being used) and it bothered me lol

barkincavdaroglu:
Since the algorithm is probabilistic, we won't always find the global mincut. For example, for the test case, we get approx. 91% success rate for finding global mincut. And some other graphs might have smaller success rates. In any case, even if the algorithm doesn't always find the global mincut, it does find the next better mincut.

src-tauri/src/globalmincut.rs
ajmcquilkin: Oh okay fair enough, let's keep an eye on that then into the future and make sure to remove them once they're not needed

PR #69, opened 2022-11-02
Fixed module hierarchy

barkincavdaroglu:
This PR separates files into their relevant directories. I also separated Edge and Node structures into their own files for a cleaner code.


ajmcquilkin (approved):
Very nice, great change! Assuming this compiles haha, haven't set up actions yet

PR #79, opened 2022-11-03
Implemented Stoer-Wagner Algorithm

### LINKED ISSUE #75: Implement Stoer–Wagner algorithm
The algorithm uses the following components:

- Union-Find
- Priority Queue (will use Fibonacci Heap for best amortized runtimes)

The algorithm is explained [here](https://dl.acm.org/doi/pdf/10.1145/263867.263872).

We might prefer this algorithm over Karger-Stein because it's not probabilistic: it always finds the exact solution.

barkincavdaroglu:
As I have discussed [here](https://github.com/ajmcquilkin/Meshtastic-emergency-response-client/pull/62), the previous algorithm, Karger-Stein algorithm, is a probabilistic one. In ~90% of cases it is expected to find the global mincut. However, Stoer-Wagner is an exact and faster algorithm. It finds the global mincut in all cases, and runs much faster. 

This PR implements Stoer-Wagner and adds the following modules:

- Binary Heap
- Union-Find
- Cut
- Stoer-Wagner Graph (enhances Graph by containing Union-Find and contract_edge functionality)

More information about the algorithm can be found [here](https://dl.acm.org/doi/pdf/10.1145/263867.263872).

Closes #75

### DIFF
--- a/src-tauri/src/algorithms/articulation_point.rs
+++ b/src-tauri/src/algorithms/articulation_point.rs
@@ -1,3 +1,6 @@
+#![allow(dead_code)]
+#![allow(non_snake_case)]
+
 use crate::graph::graph_ds::Graph;
 use core::cmp::min;
 use defaultdict::DefaultHashMap;
--- a/src-tauri/src/algorithms/stoer_wagner.rs
+++ b/src-tauri/src/algorithms/stoer_wagner.rs
@@ -1,0 +1,114 @@
+#![allow(dead_code)]
+#![allow(non_snake_case)]
+
+use crate::aux_data_structures::{
+    binary_heap::BinaryHeap, cut::Cut, stoer_wagner_ds::StoerWagnerGraph,
+};
+
+pub fn stMinCut(G: &mut StoerWagnerGraph) -> Option<Cut> {
+    let mut bheap = BinaryHeap::new(G);
+    //bheap.build_heap();
+
+    let uncontracted = G.uncontracted.clone();
+    let mut S = Vec::new();
+
+    while S.len() < uncontracted.clone().len() {
+        let max = bheap.extract_max().unwrap();
+        S.push(max.clone());
+
+        for node_id in G.uncontracted.clone() {
+            let w = G
+                .graph
+                .get_edge_weight(node_id.clone(), max.node.name.clone(), None, None);
+            bheap.update(node_id.clone(), w);
+        }
+        bheap.build_heap();
+    }
+
+    let weight = S[S.len() - 1].weight;
+    if S.len() < 2 {
+        return None;
+    }
+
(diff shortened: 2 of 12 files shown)


src-tauri/src/algorithms/articulation_point.rs
```diff
@@ -1,3 +1,6 @@
+#![allow(dead_code)]
+#![allow(non_snake_case)]
```
ajmcquilkin, line 2: Nit: I feel like it would be worth it to keep this case check, what is causing those warnings?

src-tauri/src/algorithms/stoer_wagner.rs
```diff
@@ -0,0 +1,113 @@
...
+            let counter = counter_hashmap
+                .entry(mincut.get_weight() as i32)
+                .or_insert(0);
+            *counter += 1;
+        }
+
+        println!("Counter hashmap: {:?}", counter_hashmap);
+
+        assert_eq!(counter_hashmap.len(), 1);
+        assert_eq!(counter_hashmap.get(&1), Some(&100));
```
ajmcquilkin, line 111: What specifically do these assert statements test?

src-tauri/src/aux_data_structures/cut.rs
```diff
@@ -0,0 +1,23 @@
+pub struct Cut {
+    weight: f64,
+    a: String,
+    b: String,
```
ajmcquilkin, line 4: Would it be worth making this generic? Idk what other values of `a` or `b` could be, or are they always going to be strings?

ajmcquilkin (approved):
Looks solid! Just a few minor comments

src-tauri/src/algorithms/stoer_wagner.rs
barkincavdaroglu: So counter_hashmap counts how many times a mincut weight occurs after running the algo 100 times. Since this is an exact algorithm, we want there to be one key in the map and that its value be 1 (the mincut weight).

src-tauri/src/aux_data_structures/cut.rs
barkincavdaroglu: Yeah a and b are the names of the nodes. They are the roots of their respective disjoint subsets.

src-tauri/src/algorithms/articulation_point.rs
barkincavdaroglu: Personally I hate having unnecessary warnings in my code. My class function names are snake case and algo function names are camel case (functions that don't belong to a class).

barkincavdaroglu: No actually I just name my functions randomly lol

ajmcquilkin (approved):

PR #80, opened 2022-11-04
Added function to recover the mincut

barkincavdaroglu:
After running Stoer-Wagner, we recover the mincut subsets as follows:
```rust
let nodes = vec![u, v, w, x, y, z, a, b];
let (s, t) = recover_mincut(graph_sw, nodes);
```

The tuple (s, t) is a tuple of Vec each of which contains the disjoint subset representing mincut. We pass in the nodes vector because the graph in graph_sw has been modified (all nodes except s-t are deleted due to edge contraction).

### DIFF
--- a/src-tauri/src/algorithms/stoer_wagner.rs
+++ b/src-tauri/src/algorithms/stoer_wagner.rs
@@ -1,5 +1,7 @@
 #![allow(dead_code)]
 
+use std::collections::HashMap;
+
 use crate::aux_data_structures::{
     binary_heap::BinaryHeap, cut::Cut, stoer_wagner_ds::StoerWagnerGraph,
 };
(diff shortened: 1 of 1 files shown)


ajmcquilkin:
Make sure to connect your ZenHub issues :)

src-tauri/src/algorithms/stoer_wagner.rs
```diff
@@ -93,6 +118,15 @@ mod tests {
...
 
+        let graph_sw = &mut StoerWagnerGraph::new(g.clone());
+        let _mincut = stoer_wagner(graph_sw);
+
+        let nodes = vec![u, v, w, x, y, z, a, b];
+        let (s, t) = recover_mincut(graph_sw, nodes);
+
+        println!("s: {:?}", s);
+        println!("t: {:?}", t);
+
```
ajmcquilkin, line 129: Can we create a test that validates this mincut on a known graph and validates the received cuts are accurate?

ajmcquilkin (approved):

src-tauri/src/algorithms/stoer_wagner.rs
barkincavdaroglu: Not necessarily. At least not with the example graph because there are several global mincuts. We can't know which one the algorithm returns. I will add an issue to track this though. Might be a good idea to create a graph which has a single mincut.

PR #82, opened 2022-11-04
Added module to merge and normalize edge weights

### LINKED ISSUE #81: Merge distance and radio signal quality features
We have a vector of distance values, D, and a vector of radio signal qualities, R:

For every x in D: x = (x - min(D)) / (max(D) - min(D))
For every x in R: x = (x - min(R)) / (max(R) - min(R))

Now all values in both D and R are within range (0, 1). The edge weights are then:
weights = (0.5 x D) + (0.5 x R)

barkincavdaroglu:
We have a vector of distance values, D, and a vector of radio signal qualities, R:

For every x in D: x = (x - min(D)) / (max(D) - min(D))
For every x in R: x = (x - min(R)) / (max(R) - min(R))

Now all values in both D and R are within range (0, 1). The edge weights are then:
weights = (0.5 x D) + (0.5 x R).

We use the module as follows:
```rust
let u = NodeIndex::new(0);
let v = NodeIndex::new(1);
let w = NodeIndex::new(2);
let x = NodeIndex::new(3);

let a = vec![u.clone(), u.clone(), w.clone(), v.clone()];
let b = vec![v.clone(), w.clone(), x.clone(), x.clone()];

let distance = vec![0.45, 0.67, 0.23, 1.2];
let radio_s_quality = vec![5.5, 3.12, 10.3, 2.7];

let edges = edge_factory(a, b, distance, radio_s_quality);
```

Closes #81

### DIFF
--- a/src-tauri/src/algorithms/stoer_wagner.rs
+++ b/src-tauri/src/algorithms/stoer_wagner.rs
@@ -124,8 +124,8 @@ mod tests {
         let nodes = vec![u, v, w, x, y, z, a, b];
         let (s, t) = recover_mincut(graph_sw, nodes);
 
-        println!("s: {:?}", s);
-        println!("t: {:?}", t);
+        assert!(s.len() + t.len() == 8);
+        assert!(s.iter().all(|x| !t.contains(x)));
 
         for _ in 0..100 {
             let graph_sw = &mut StoerWagnerGraph::new(g.clone());
--- a/src-tauri/src/algorithms/stoer_wagner.rs
+++ b/src-tauri/src/algorithms/stoer_wagner.rs
@@ -1,5 +1,7 @@
 #![allow(dead_code)]
 
+use std::collections::HashMap;
+
 use crate::aux_data_structures::{
     binary_heap::BinaryHeap, cut::Cut, stoer_wagner_ds::StoerWagnerGraph,
 };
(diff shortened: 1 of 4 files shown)


src-tauri/src/algorithms/stoer_wagner.rs
```diff
@@ -49,6 +51,29 @@ pub fn stoer_wagner(graph: &mut StoerWagnerGraph) -> Cut {
     }
 }
 
+pub fn recover_mincut(
```
ajmcquilkin, line 54: How does this differ from #80?

src-tauri/src/algorithms/stoer_wagner.rs
```diff
@@ -93,6 +118,15 @@ mod tests {
 
         let mut correct_mincut_weight_count = 0;
 
+        let graph_sw = &mut StoerWagnerGraph::new(g.clone());
+        let _mincut = stoer_wagner(graph_sw);
+
+        let nodes = vec![u, v, w, x, y, z, a, b];
+        let (s, t) = recover_mincut(graph_sw, nodes);
+
+        println!("s: {:?}", s);
```
ajmcquilkin, line 127: Might want to put this behind a non-production (debug) flag, we don't want to be logging in prod

src-tauri/src/aux_functions/edge_factory.rs
```diff
@@ -0,0 +1,86 @@
+use crate::graph::edge::Edge;
+use petgraph::graph::NodeIndex;
+
+pub fn edge_factory(
+    a: Vec<NodeIndex>,
+    b: Vec<NodeIndex>,
+    distance: Vec<f64>,
+    radio_s_quality: Vec<f64>,
+) -> Vec<Edge> {
+    let distance_min = distance.iter().fold(f64::INFINITY, |acc, &x| acc.min(x));
```
ajmcquilkin, line 10: Is the purpose of this to simply create edges? Might be worth an inline comment

src-tauri/src/aux_functions/mod.rs
```diff
@@ -0,0 +1 @@
+pub(crate) mod edge_factory;
```
ajmcquilkin, line 1: What is the `(create)` symbol out of curiosity?

ajmcquilkin (approved):
Looks solid! Few doc comments

src-tauri/src/algorithms/stoer_wagner.rs
barkincavdaroglu: Off for some reason the changes from previous branch leaked here? They don't differ.

src-tauri/src/aux_functions/mod.rs
barkincavdaroglu: Not super sure vscode recommended that I put that.

ajmcquilkin (approved):

PR #91, opened 2022-11-04
Display All Nodes on Map

### LINKED ISSUE #89: Display All Nodes on Map
Currently, the web client only displays a single node on the map (the node connected over serial). We need to store which nodes we have heard from (and when), and display information on these nodes on the map.

ajmcquilkin:
This PR refactors the redux node handling flow to allow for the rendering of all nodes within the `MapView` component. This change also allows for multiple nodes to be simultaneously connected to the client computer.

image

Closes #89


barkincavdaroglu (approved):
lgtm

PR #93, opened 2022-11-04
Add another scaler to radio_signal_quality for a better scaling of the features

barkincavdaroglu:
High radio signal quality will usually come with low distance, so just scaling the features with minmax will negate the other's magnitude. This PR scales the radio_signal_quality with 1/ln(1+x) before minmax scaling so that large values of radio_signal_quality will result in small values and vice versa.


barkincavdaroglu:
We add 0.001 in minmax scaler to make sure no weight is 0.0

ajmcquilkin (approved):
Looks solid!

PR #99, opened 2022-11-06
[to be discarded] Add Protobuf to Graph Dataflow

uhuruhashimoto:
**The Neighbor Problem**

This pull request addresses the problem of centralizing a decentralized mesh network. While we could (and do) append route info to packets, this would provide us with knowledge of only the shortest paths in the network. In a rescue situation, where reliability is paramount, we need a better protocol. We can break this down into two steps: 

1. Neighbor discovery
2. Transmission to coordinator

For the first step, all nodes in the network need to send out a simple packet to their neighbors (1 hop away) to establish connection. If neighbor lists are known, we skip this step and send info directly to the coordinator. 

This PR assumes that we get info in the form of N protobufs, where N is the number of connected nodes in the network. Each protobuf contains the neighbor info for each node - IDs and positions. We start with that and use it to create the edges on our graph. 

(This information could also be encoded in regular packet form, removing the need for a new protobuf. If so, it's easy enough to delete our `neighbors.proto` file.)


ajmcquilkin:
Is there a reason this PR is still open with the other two?

uhuruhashimoto:
> 

No; closed

Issue #43, opened 2022-10-24
Implement Infrastructure for Running Generic Algorithms on Stored Graph


ajmcquilkin:
Allow developers to run any generic graph-based algorithm on the stored graph.

PR #107, opened 2022-11-08
Graph Initialization

### LINKED ISSUE #40: Create Infrastructure for Initializing and Storing Graph Data Structure
- Add new redux saga for handling application initialization
- Add a node TS class with MAC address as primary key
- Add global application graph structure
- Implement initialization of graph based on a given set of nodes
- Implement functionality for adding a node to the graph

### LINKED ISSUE #65: Connect Data to Graph
In this ticket, we translate mocked data into a graph, and store that graph in a centralized location (initially without updates - this is just the data structure initialization flow).

uhuruhashimoto:
This PR sets up the dataflow for basic graph initialization, using a struct of protobufs to assemble the necessary neighbor data for nodes. It leaves stubs in the code for frontend connections and data conversion. 

Putting this PR up partly for feedback on the process itself. Because of that, here's a list of questions I have. 

Fixed: 
1.) Distance normalization: this results in infinite weights in the case of identical minimum/maximum distance values. Fixed by adding epsilon. 

**Note on Data Conversion**
different algorithms return different datasets, though we provide them all the same graph. The graph->return flow will be addressed in a follow-up PR. 

Closes #40 and #43
Closes #65 (a sub-issue of 40)


barkincavdaroglu:
We almost never have the same minimum and maximum, but to handle such edge cases it would suffice to add epsilon to the maximum so that they are never equal.

src-tauri/src/aux_functions/datatypes.rs
```diff
@@ -0,0 +1,27 @@
...
+* This is the info we're given for each node, which records (in
+* protobuf structs) its user and position info and user/position info for each neighbor.
+*
+* We'll recieve N of these and use them to create a graph with N nodes.
+*/
+pub struct NeighborInfo {
+    pub(crate) user: User,
+    pub(crate) position: Position,
+    pub(crate) num_neighbors: u32,
+    pub(crate) neighbors: Vec<Neighbor>,
```
barkincavdaroglu, line 13: Why do you have pub(crate) for these? pub(crate) creates a public module so that other files can use it. Is there a module for num_neighbors or neighbors?

src-tauri/src/aux_functions/datatypes.rs
```diff
@@ -0,0 +1,27 @@
...
+}
+
+pub struct Neighbor {
+    pub(crate) user: User,
+    pub(crate) position: Position,
```
barkincavdaroglu, line 18: Same here. I think they should just be pub?

src-tauri/src/aux_functions/graph_init.rs
```diff
@@ -0,0 +1,268 @@
...
+* Do postprocessing of the algorithm result and return it in the format the frontend expects.
+*/
+pub fn convert_data_for_display(graph: Graph) {
+    //TODO:
+}
+
+/*
+* This function creates a graph from the input data.
+*/
+pub fn load_graph(data: Vec<NeighborInfo>) -> Graph {
```
barkincavdaroglu, line 25: Looks good!

src-tauri/src/aux_functions/datatypes.rs
uhuruhashimoto: Ditto; fixed

barkincavdaroglu (approved):

barkincavdaroglu (commented):
Oh I see you included the code in this PR in #124. Can you close this pr?

PR #108, opened 2022-11-08
add epsilon while scaling weight to take care of the edge case where min = max

barkincavdaroglu:
In response to #107


ajmcquilkin (approved):
Looks good, but might be worth explaining the problem a bit more in the PR or connecting an issue to the PR

PR #114, opened 2022-11-09
Implement Dummy Region and Layer UI

### LINKED ISSUE #37: Create Dummy Region and Layers Buttons
Implement non-functional buttons for layers and regions.

ajmcquilkin:
This PR implements **non-functional** UI for future region and layer functionality, as shown below. Both buttons will trigger an alert when clicked stating the UI is currently non-functional.

image

Closes #37


sherry1iu (approved):
lgtm

PR #124, opened 2022-11-12
Run Algorithms on Demo Data from the Frontend

### LINKED ISSUE #43: Implement Infrastructure for Running Generic Algorithms on Stored Graph

### LINKED ISSUE #64: Mock Protobuf Data
To run analysis on graphs, we need to mock protobuf data. This involves deciding which fields to use and possibly generating a series of random protobufs. 

(This may also run into the neighbor problem - if the edges on our graph are connections, then nodes need to know  their neighbors and update the coordinator if neighbors change. If necessary, mock new protobuf fields for this change).

### LINKED ISSUE #66: Send Data to Frontend
This issue involves creating a Tauri event to send updated data (graph analysis results) to the frontend.

### LINKED ISSUE #123: Non-corresponding member definitions between Rust structs/Tauri interfaces
According to Tauri docs, passing data between the front and the backend requires it to be serde serializable (https://tauri.app/v1/guides/features/command). Our compiled rust structs (protobufs) weren't, so I used prost-wk types to add serde serialization at compile time. I'm now accessing the js/ts definition of protobuf interfaces in the frontend and the rs version in the backend, and seems like everything should be peachy.

However, I'm running into another problem with struct members. When we compile protobufs to ts/js interfaces, it names the members in camelCase. For example, NodeInfo has members named, respectively, lastHeard in @meshtasticjs and last_heard in app::protobufs.This means that the frontend and backend don't line up in their struct definitions, and we can't pass data between them.

We haven't run into this before because we haven't tried to convert between ts interfaces and rs structs -- we've just been using the ts directly to generate markers in the frontend. I see a couple of solutions:
1.) Using the ts-rs crate to compile protos from typescript (instead of compiling them to rust with prost).
2.) Manually using an intermediate struct I defined (NeighborInfo) to pull data from protobufs and send it back. This wouldn't add too much overhead, since we only need one type of packet, but it's not extensible to data flow generally between front and backend. Of course, if we get data from serial, this won't be a problem forever.

uhuruhashimoto:
Summary (See Issue #123):

When passing data from the frontend to the backend, we encounter two issues: 

1.) All types need to be serde serializable, requiring us to use the well-known types feature of the `prost` crate to append serialization at compile-time. This also requires the protoc compiler to be installed - `protoc_bin_vendored` only works at build time. 

2.) When we create structs, we compile C++ to Typescript interfaces and to Rust structs. The members of these structs use different naming conventions; that is, `NodeInfo` in `@meshtasticjs` implements different members than `NodeInfo` of `app::protobufs`. Because of this, we can't simply pass protobuf packets from the frontend to the backend. 

Possible solutions to 2 include: 
a.) using a `ts-rs` crate to compile Rust protobufs from TypeScript, and 
b.) using an intermediate struct to pull necessary data from the protobufs and send that to the backend, and (3) specifying member case at compile time. 

This PR implements the approach b. In it, we store necessary info from protobufs in NeighborInfo structs (of which there are both Rust and TS analogs), which gives us finer control over data serialization and contents. We pass this to the backend, run algorithms on it, and return the result to the console as a string. 

Testing:
- unit testing added for `load_graph`
- `pnpm run rust:dev` compiles and runs; button clicks produce desired console/terminal output with no errors. 

Notes:
- This could easily be configured to run with actual data, but we would need to make assumptions about which neighbors nodes have (likely assuming that the graph is fully connected). 

Next Steps:
Expand demo dataset to test algorithms more thoroughly.

Closes #123 
Closes #43 
Closes #66 
Closes #64


barkincavdaroglu:
Could you add Serialize and Deserialize traits for our graph too?

src-tauri/src/aux_functions/graph_init.rs
```diff
@@ -0,0 +1,143 @@
...
+            neighbor_info_2,
+            neighbor_info_3,
+            neighbor_info_4,
+        ]);
+        // Check that the graph has the correct number of nodes
+        assert_eq!(graph.get_order(), 4);
+        // Check that the graph has the correct number of edges
+        assert_eq!(graph.get_size(), 12);
+    }
+}
```
barkincavdaroglu, line 143: Perfect!

src-tauri/src/main.rs
```diff
@@ -74,3 +87,42 @@ async fn async_process_model(
 
     Ok(())
 }
+
+#[tauri::command]
+fn run_articulation_point(nodes: Vec<NeighborInfo>) -> String {
```
barkincavdaroglu, line 92: Can you move these functions to a new file? They probably shouldn't be in main.rs

barkincavdaroglu (changes requested):

src-tauri/src/aux_functions/graph_init.rs
uhuruhashimoto: :)

src-tauri/src/main.rs
uhuruhashimoto: done; moved to aux_functions

uhuruhashimoto:
> Could you add Serialize and Deserialize traits for our graph too?

Because petgraph doesn't directly implement serde serialization (see https://github.com/petgraph/petgraph/issues/139), I can't just add the `#[derive (Serialize, Deserialize)]` to the struct like I can to a struct with serializable members. Working on how to add that manually

ajmcquilkin:
Re: edit, do the protobufs not automatically implement the serde macros? If I'm correct, the compiled protobufs have native support for serialization and deserialization

uhuruhashimoto:
> Re: edit, do the protobufs not automatically implement the serde macros? If I'm correct, the compiled protobufs have native support for serialization and deserialization. Not sure offhand how to handle the graph serialization issue

They do not; unless we specifically add it at compile time. I'm not doing that right now because the solution I found required protoc, but it's easy enough to add. Because we're not passing the structs themselves yet, it's not yet necessary as of this PR. 
We can get graph serialization with serde-1; I'll add that (and testing for it) as ticket #126

barkincavdaroglu (approved):

PR #127, opened 2022-11-14
Run Algorithms on Live Data

uhuruhashimoto:
This PR makes the three-line change in the Sidebar to pass actual node data to the backend. 

It also implements corresponding null checks to make sure that the app does not panic if there are no nodes detected. Unit tests are updated accordingly.

As of this PR, we make the assumption that our graph is fully connected, or at least that neighbors are known, by artificially adding them to neighbor lists in `generateNeighborInfo` in `DemoData.tsx`. This is a temporary assumption and can easily be changed.

Testing: 
- tested by running `pnpm run rust:dev` with no nodes

### DIFF
--- a/src-tauri/src/algorithms/stoer_wagner.rs
+++ b/src-tauri/src/algorithms/stoer_wagner.rs
@@ -24,11 +24,10 @@ pub fn st_mincut(g: &mut StoerWagnerGraph) -> Option<Cut> {
         }
         bheap.build_heap();
     }
-
-    let weight = a[a.len() - 1].weight;
     if a.len() < 2 {
         return None;
     }
+    let weight = a[a.len() - 1].weight;
 
     let s = a[a.len() - 1].node.name.clone();
     let t = a[a.len() - 2].node.name.clone();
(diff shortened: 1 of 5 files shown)


src-tauri/src/algorithms/stoer_wagner.rs
```diff
@@ -24,29 +24,42 @@ pub fn st_mincut(g: &mut StoerWagnerGraph) -> Option<Cut> {
...
+    let weight = a[a.len() - 1].weight;
 
     let s = a[a.len() - 1].node.name.clone();
     let t = a[a.len() - 2].node.name.clone();
 
     return Some(Cut::new(weight, s, t));
 }
 
-pub fn stoer_wagner(graph: &mut StoerWagnerGraph) -> Cut {
+pub fn stoer_wagner(graph: &mut StoerWagnerGraph) -> Option<Cut> {
```
barkincavdaroglu, line 38: Why the changes?

uhuruhashimoto: Thanks for asking! :) If there are no nodes in the graph, the app crashes because the cut returns None (see `stoer_wagner` function; in stoer_wagner.rs, line 61). I wrapped this both functions in an Option so we don't panic if we unwrap that value, and updated the unit test to take an Option instead of a Cut. 

So there are no changes to the return logic of `stoer_wagner` itself, just to the way we're handling its return value.

uhuruhashimoto: Could also be handled with an `unwrap_or` with reasonable defaults, but I didn't think it was worth it to make up a cut (especially since we prioritize cuts by weight, so we'd need to set a high weight). That felt contrived, so I used `Option` instead.

barkincavdaroglu (approved):

PR #151, opened 2022-12-08
Add Working Serial Read and Write

### LINKED ISSUE #88: Implement MVP Channel Communication
Implement basic channel communication, where a user on the client can send and receive notifications within the UI.

ajmcquilkin:
image

This PR introduces working serial communication (read + write) that allows the client to communicate with a connected serial device. Testing of this change can be done as follows:

1. Run `pnpm run rust:dev` and open the web console in the resulting window (right click -> "Inspect Element" -> Console)
2. Run `__TAURI_INVOKE__("get_all_serial_ports").then(console.log).catch(console.error)`. This command will then list all serial ports the application is able to access. Determine which port is connected to your node
3. Run `__TAURI_INVOKE__("connect_to_serial_port", { portName: "{PORT_NAME}" }).then(console.log).catch(console.error)`, where `{PORT_NAME}` is replaced with the name of the port you found in the previous step. This will print `true` in the console if it is successful, and `false` otherwise. A return value of `false` indicates an error occurred in the connection process. Check the Rust console for more information
4. All packets your node receives will now be displayed in the client console
5. To send text data, run `__TAURI_INVOKE__("send_text", { text: "{TEXT_MESSAGE}" }).then(console.log).catch(console.error)` in your client console, where `{TEXT_MESSAGE}` is the string message you want to send over the mesh

This change works by building a `SerialConnection` struct, which manages the details of the serial connection to the connected node. An instance of this connection struct is stored in [Tauri managed state](https://gist.github.com/captainhusaynpenguin/5bdb6fcb141628b6865619bcd1c827fd), and can be modified through [Tauri commands](https://tauri.app/v1/guides/features/command/). This allows the connection to persist across requests to the backend. Packet transmission from Rust to TS is handled with [Tauri events](https://tauri.app/v1/guides/features/events/).

Closes #88


barkincavdaroglu (approved):
👁

src-tauri/src/main.rs
```diff
@@ -1,80 +1,116 @@
...
+    match connection.connect(port_name, 115_200) {
+        Ok(_h) => (),
+        Err(_e) => {
+            eprintln!("Could not connect to radio");
+            return false;
+        }
+    };
+
+    let mut decoded_listener = match connection.on_decoded_packet.as_ref() {
+        Some(l) => l.resubscribe(),
```
barkincavdaroglu, line 61: do these variable names represent anything or are they random?

barkincavdaroglu: Oh okay you are using the first letter

ajmcquilkin: Yeah, happy to change that if it doesn't make sense or isn't clear

barkincavdaroglu: all good!

PR #153, opened 2022-12-17
Handle Disconnection and Save Data to Local DB

barkincavdaroglu:
This PR has a data structure Timeline that will be initialized for each rescue attempt to write to local database using PostgreSQL. We use 2 tables:
1. training_table has 2 columns, [timeline_id, label] where label represents whether the timeline has failed or not (0 when fails, 1 otherwise),
2. ft_table (ref by timeline_ft_table_name) has 5 columns, [timeline_id, snapshot_id, node_fts_prim, snapshot_txt, misc]. node_fts_prim is json object with node_id mapping to its features. As an example, 
```
{
    "a": {
        "speed":0,
        "degree":0.9,
        "altitude":1,
        "latitude":43.71489,
        "direction":0,
        "longitude":-72.28486
    },
    "b": {
        "speed":0,
        "degree":1.51,
        "altitude":1,
        "latitude":43.71584,
        "direction":0,
        "longitude":-72.28239
    }
}
```
snapshot_txt is the one that take_snapshot_of_graph returns, and misc is json field for other information we may want to use, such as global minimum cut or articulation points.

The Timeline contains the most current snapshot. Each time a new snapshot is taken, it checks whether a disconnection has occurred or not using the difference in number of nodes between the snapshot timeline contains and the snapshot that is passed into add_snapshot function (of Timeline). If there is a disconnection, we end the current timeline, label it as 0 in training table, and start a new timeline whose first snapshot is the function parameter.

**The test expects a connection to postgresql database with the given config, either install postgresql or comment the test block.**


barkincavdaroglu:
Actually let's put this in the backlog for now.

ajmcquilkin:
Do you mean to have this as draft? I noticed you requested a review

barkincavdaroglu:
> Do you mean to have this as draft? I noticed you requested a review

Yes. As can be seen above, I have (1) converted the PR to draft and (2) left a comment saying this should go into the backlog.

ajmcquilkin:
Oop sorry, somehow missed that!

ajmcquilkin:
Is this still a PR that's mergeable and/or relevant? If not let's close it

PR #155, opened 2022-12-29
Add Loading and Serial Connection UI

ajmcquilkin:
This PR adds in "Application Loading" UI, as well as a page-based routing system ([see here](https://github.com/remix-run/react-router)) with a new "Connect to Serial" page. This page allows for connection to and disconnection from hardware serial devices. This is shown in the video below.

https://user-images.githubusercontent.com/46639306/209896560-4fd7778d-e2ca-4d41-9496-ddbd1144b7c8.mp4

The page-based routing system replaces the existing redux-based pane management framework. Instead of maintaining an enum of all valid panes and views, the application will now use the `/view_name` route for new (major) views (e.g. map view, serial connection view), and `/view_name#subview_name` for minor views (e.g. message or node search panes within the map view). Currently, we have only the `/` (map) and `/serial-connect` (serial connection) views, where each pane within the map view is controlled by a [hash](https://developer.mozilla.org/en-US/docs/Web/API/Location/hash).

Additionally, this PR adds a standard `pnpm run rust:build` command, which builds the application for use on the developer's native OS.


sherry1iu:
The port on my computer isn't being detected

ajmcquilkin:
Interesting, just so we're on the same page do you have a device connected to your computer that is on with somewhat recent firmware?

sherry1iu:
Yes, I flashed it last week

ajmcquilkin:
Hmm interesting, does the app start correctly? If so, could you send a screenshot of your terminal?

sherry1iu:
> Hmm interesting, does the app start correctly? If so, could you send a screenshot of your terminal?

image

I'm using pnpm run ui:dev

ajmcquilkin:
Oh the `ui:dev` command won't work anymore since we're now using Rust serial ports, try running `pnpm run rust:dev`

barkincavdaroglu (approved):

ajmcquilkin:
Tagging for review @timofei7

PR #154, opened 2022-12-26
Added Rust device state and functioning broadcast communications

ajmcquilkin:
This PR adds a new state-management paradigm within the application. Previously all application state was held within the TS layer, which made meaningful data processing and algorithmic interpretation challenging in the Rust layer. This PR adds a new `MeshDevice` struct in the Rust management layer that will hold all state for each device connected via serial, and exposes methods to update its state when packets are received from the radio. When a new packet comes in that updates the `MeshDevice`'s state, the backend can choose to dispatch an updated state to the client, which will send a complete copy of the `MeshDevice` state to the client. This is a known inefficiency, but will vastly simplify development for the time being. This inefficiency is acceptable in the short-term because of how infrequently this whole-state transmission would occur (either on a received telemetry packet or on a sent or received message).

image

Additionally, this PR implements working communication through the broadcast channel (id 0). The UI requests that the backend send a text message, which will then update the `MeshDevice` state and dispatch the updated state to the client.

> Note that all files changed within the `/src-tauri/bindings` directory are auto-generated from backend Rust types for consumption, and are not manually coded. This generation runs automatically on `pnpm run rust:test`. For more information, see [here](https://github.com/Aleph-Alpha/ts-rs).

Closes #110 
Closes #84

PR #156, opened 2022-12-29
Revert "Added Rust device state and functioning broadcast communications"

sherry1iu:
Reverts ajmcquilkin/Meshtastic-emergency-response-client#154

After the PR, serial connection isn't working for me anymore - there is no longer a pop-up when clicking the link icon

image


ajmcquilkin:
Yeah try cloning the new PR locally and lmk if that works haha

PR #159, opened 2023-01-06
Uhuruhashimoto/simplify api

uhuruhashimoto:
## Objective 
Remove unnecessary structures previously used to pass mock data to the backend from the frontend.

## Implementation
Negative-code PR that removes placeholders instead of commenting them. While we'll eventually need some representation of a graph/algo results in the frontend, we'll keep the codebase cleaner by building data structures for it specifically instead of repurposing these. 
Note: left NeighborInfo struct on the backend intact, as it can be modified and used to mock data immediately (as per meeting). A follow-up PR should generate mock protobufs for testing purposes. 

## Testing
`pnpm run ui:dev`
`pnpm run rust:dev`

## Tickets
#2


ajmcquilkin (approved):
Looks great! Ran and didn't see any regressions

PR #161, opened 2023-01-08
Fix Windows serial connection bug

ajmcquilkin:
This PR fixes a bug in which users on Windows could not connect to serial ports, getting the message "Access is denied" when attempting to connect. This PR resolves this bug, as well as a visual bug on Windows where the `Messages` view will always display a scrollbar.


sherry1iu (approved):
This works! Tysm :)

ajmcquilkin:
Awesome, glad to hear it!

PR #162, opened 2023-01-08
Uhuruhashimoto/protobuf conversion

uhuruhashimoto:
## Goal
Centralize protobuf conversion constants into one file and use it in graph construction. 

## Implementation
Lat/lng constants are public in `conversion_constants.rs`. We may want to organize them as we add more for different protobuf fields so we don't pollute namespaces more than necessary (for example, structs by category). 

## Testing
unittests pass.

### DIFF
--- a/src-tauri/src/aux_functions/graph_init.rs
+++ b/src-tauri/src/aux_functions/graph_init.rs
@@ -4,6 +4,26 @@ use crate::aux_data_structures::neighbor_info::{Neighbor, NeighborInfo};
 use crate::graph::graph_ds::Graph;
 use petgraph::graph::NodeIndex;
 
+/* Lat: 1e-7 conversion from int to floating point degrees; see mesh.proto */
+const LAT_CONVERSION_FACTOR: f64 = 1e-7;
+/* Longitude: 1e-7 conversion from int to floating point degrees; see mesh.proto */
+const LON_CONVERSION_FACTOR: f64 = 1e-7;
+/* Altitude: in meters above sea level, no conversion needed */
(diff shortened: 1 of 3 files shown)


src-tauri/src/aux_functions/graph_init.rs
```diff
@@ -61,20 +62,15 @@ pub fn load_graph(data: Vec<NeighborInfo>) -> Graph {
...
 fn calculate_converted_distance(x: f64, y: f64, z: f64, nbr_x: f64, nbr_y: f64, nbr_z: f64) -> f64 {
-    let conversion_factor = (10.0 as f64).powi(-7);
     return total_distance(
-        (x as f64) * conversion_factor,
-        (y as f64) * conversion_factor,
-        z as f64,
-        (nbr_x as f64) * conversion_factor,
-        (nbr_y as f64) * conversion_factor,
-        nbr_z as f64,
+        (x as f64) * LAT_CONVERSION_FACTOR,
```
ajmcquilkin, line 68: What is the purpose of the casting here? I can't see a reason this cast is necessary

ajmcquilkin (approved):
Looks good! Just a minor thought

src-tauri/src/aux_functions/graph_init.rs
barkincavdaroglu: Agreed.

uhuruhashimoto: Probably a holdover from something in the distant past. Removed all casts from the wrapper, thanks for the feedback! :)

ajmcquilkin (approved):

PR #173, opened 2023-01-09
Implemented UI for settings

### LINKED ISSUE #101: Implement settings screen according to UI
Set up initial settings page - no specific settings

sherry1iu:
This PR implements the UI for the user configuration settings, namely device ID, device nickname, and device name; as well as a save button.

image

Closes #101

### DIFF
--- a/src/components/Settings/Settings.tsx
+++ b/src/components/Settings/Settings.tsx
@@ -3,18 +3,17 @@ import { useNavigate } from "react-router-dom";
 import { XMarkIcon } from "@heroicons/react/24/solid";
 import { Input } from "@material-tailwind/react";
 
-function Settings() {
+const Settings = () => {
   const navigateTo = useNavigate();
 
   const [deviceID, setDeviceID] = useState("Default Device ID");
-
   const [deviceName, setDeviceName] = useState("Mr. Meshtastic");
   const [deviceNickname, setDeviceNickname] = useState("MRM");
 
   const handleSubmit = (e) => {
     e.preventDefault();
-    const sherryTest = { deviceID, deviceNickname, deviceName };
-    console.log(sherryTest);
+    const deviceInfoTest = { deviceID, deviceNickname, deviceName };
+    console.log(deviceInfoTest);
   };
 
   return (
@@ -37,6 +36,9 @@ function Settings() {
           <div className="flex justify-center sticky top-0 z-50">
             <div className="font-bold"> Settings </div>
           </div>
+        </div>
+        <div className="text-sm flex justify-center">
+          Note: This is the UI, and is not currently connected to backend
         </div>
 
         {/* Make a form where we put the options */}
@@ -109,6 +111,6 @@ function Settings() {
       </div>
     </div>
   );
-}
+};
 
 export default Settings;
--- a/src/components/Settings/Settings.tsx
+++ b/src/components/Settings/Settings.tsx
@@ -86,18 +86,7 @@ const Settings = () => {
                 onChange={(e) => setDeviceNickname(e.target.value)}
               />
             </div>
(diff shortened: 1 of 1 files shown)


ajmcquilkin:
Can we somehow display that this settings page is non-functional? Maybe some text in the modal, maybe commenting out the settings icon temporarily, just so we don't get issues listing this as a bug

src/components/Settings/Settings.tsx
```diff
@@ -1,11 +1,22 @@
-import React from "react";
+import React, { useState } from "react";
 import { useNavigate } from "react-router-dom";
 import { XMarkIcon } from "@heroicons/react/24/solid";
 import { Input } from "@material-tailwind/react";
 
 function Settings() {
```
ajmcquilkin, line 6: Nit: Code style, the typical way we've been doing components is `const Settings = () => { ... }`

src/components/Settings/Settings.tsx
```diff
@@ -1,11 +1,22 @@
...
   const navigateTo = useNavigate();
 
+  const [deviceID, setDeviceID] = useState("Default Device ID");
```
ajmcquilkin, line 9: I would have these initialized as empty strings for the time being

src/components/Settings/Settings.tsx
```diff
@@ -1,11 +1,22 @@
...
+
+  const [deviceName, setDeviceName] = useState("Mr. Meshtastic");
+  const [deviceNickname, setDeviceNickname] = useState("MRM");
+
+  const handleSubmit = (e) => {
+    e.preventDefault();
+    const sherryTest = { deviceID, deviceNickname, deviceName };
```
ajmcquilkin, line 16: I wouldn't identify yourself in the codebase, I would recommend logging that object directly

ajmcquilkin (changes requested):
Looks solid, just a few comments (requesting changes for removing personal identifiers from the codebase)

*Edit:* also requesting changes for visible indicator that this is not currently connected to the Rust backend

src/components/Settings/Settings.tsx
```diff
@@ -15,60 +26,85 @@ function Settings() {
...
+                label="3 character nickname for small screens."
+                maxLength={3}
+                minLength={1}
+                value={deviceNickname}
+                onChange={(e) => setDeviceNickname(e.target.value)}
+              />
+            </div>
+            <label className="flex pl-[4%] pt-[6%] pl-2 text-xl">
+              {" "}
+              Option 4{" "}
```
uhuruhashimoto, line 89: What is this a placeholder for? Is it meant for a specific feature (e.g. renaming nodes)? Do we have a specific reason to include it?

src/components/Settings/Settings.tsx
uhuruhashimoto, line 16: Might want to switch the variable name `sherryTest` to `deviceInfoTest` if we're keeping the log?

sherry1iu: Meant to get rid of that, good catch!

src/components/Settings/Settings.tsx
sherry1iu: I left it since it'll eventually initialize to a string that we get from the device - I wanted to show that it was disabled, and that you can't change the box

src/components/Settings/Settings.tsx
uhuruhashimoto: Gotcha, seems fine to leave it then. I commented on it because it looked like code that could turn into a stub without any meaning. To avoid that, mind doing some combination of (1) expressing the extra options by name in the display itself, (2) documenting them in the code comments and/or (3) adding tickets to the board to finalize them?

ajmcquilkin (approved):
Looks solid!

PR #189, opened 2023-01-11
Uhuruhashimoto/mock protobufs

uhuruhashimoto:
## Goal
Setup a skeleton for making mocks so we can proceed with algo->frontend display development while changing protobufs/firmware to get graph topology.

## Implementation
Set up a mock module, currently using the fields of our NeighborInfo struct (which is subject to change if we change our protobuf fields). Also provided semi/randomly connected graph creation through shuffling the edge set. 

## Testing
Sample Output of `mock_generator(3, 0.8)` - creates a partially connected graph with 3 nodes, located in Hanover.
```
[
  NeighborInfo { 
    id: 0, 
    timestamp: 0, 
    neighbors: [
    Neighbor { 
      id: 1, 
      snr: 0.08752667779541135, 
      timestamp: 0 
    }
    ] 
  }, 
  NeighborInfo { 
    id: 1, 
    timestamp: 0, 
    neighbors: [
    Neighbor { 
      id: 0, 
      snr: 0.948595428025263, 
      timestamp: 0 
     }
    ] 
  }, 
  NeighborInfo { 
    id: 2, 
    timestamp: 0, 
    neighbors: [
      Neighbor { 
        id: 0, 
        snr: 0.1441237231664394, 
        timestamp: 0 
      }, 
      Neighbor { 
        id: 1, 
        snr: 0.5730639560523529, 
        timestamp: 0 
      }
    ] 
  }
]
```
Note: We will be using ids to get Location info from an existing hashmap. This will include latitude, longitude, altitude, and (unreliable) speed and heading information for each node. This packet exists to construct network topology specifically, and is not tied to the location update system.


ajmcquilkin:
Looks really solid! One thing I'm wondering is if it's worth maintaining the last time that each node has heard from each of its neighbors. I'm working on a design doc now for what this proto should look like, and one idea that came up was that we could include a logistic factor of time within the edge weight calculation, since if we haven't heard from a node in a while that often means the connection has dropped.

barkincavdaroglu:
Can we please tables this at least until tomorrow? I need to take a more detailed look on this. Will leave my thoughts tomorrow morning. But for now, snr is between two nodes. So node a's snr is in relation to the node b that a communicated with. If we don't have access to who that node b is (or a method to find out), then it's not really useful. And why do we have node information in neighborhood table? It would be extremely repetitive. Is there any way we can use neighborhood table to only store edge information and then have another table to store node information at the same time as neighborhood table?

src-tauri/src/mocks/mock_generator.rs
```diff
@@ -0,0 +1,74 @@
...
+pub const HANOVER_LON_PREFIX: f64 = 43.70;
+pub const HANOVER_LAT_PREFIX: f64 = 72.28;
+
+pub fn mock_generator(num_nodes: i32, percent_connected: f64) -> Vec<NeighborInfo> {
+    let mut neighborinfo_vec = Vec::new();
+    for node_id in 0..num_nodes {
+        let mut new_neighbor = NeighborInfo {
+            selfnode: Neighbor {
+                // Assign sequential ids
+                id: node_id as u32,
```
barkincavdaroglu, line 23: I think best to just have node id and snr here. And then we can have a lookup table for id --> node_fts (gps etc.)

barkincavdaroglu: We would also need speed and direction of each node

ajmcquilkin: I think the issue is that speed and direction are typically super inaccurate on the devices, not sure how much we would want to weight this, if at all

barkincavdaroglu:
> Looks really solid! One thing I'm wondering is if it's worth maintaining the last time that each node has heard from each of its neighbors. I'm working on a design doc now for what this proto should look like, and one idea that came up was that we could include a logistic factor of time within the edge weight calculation, since if we haven't heard from a node in a while that often means the connection has dropped.

@ajmcquilkin  Since we are taking snapshots each minute, we should include the latest edge. If there is, say, only one edge that occurred 59 seconds ago, then we should include that as is. We cannot make any reasonable assumption on how that weight would change with time. And at least for the ML model we would need the raw data.

src-tauri/src/mocks/mock_generator.rs
barkincavdaroglu: I believe the speed is determined by taking the difference of current and prior location. On average that's still a useful information. I am considering of using direction for something else (might or might not be useful, I will need to test, if it's not too much trouble of course).

ajmcquilkin:
> 

Yeah that's fair, although with the algorithmic analysis it would still be useful to lower the weight of an edge based on time, assuming we have an algorithm that would benefit from it (maybe probabilistic articulation point location or something). We could collect that info on each edge and then handle it differently in the ML and algos

barkincavdaroglu:
But what if they are walking towards each other? Just become time has passed does not at all mean the weight will strictly decrease.

uhuruhashimoto:
> Can we please table this (not merge) until at least tomorrow noon? I need to take a more detailed look on this. Will leave my thoughts tomorrow morning. But for now, few questions. snr is between two nodes. So node a's snr is in relation to the node b that node a communicated with. If we don't have access to who that node b is (or a method to find out), then it's not really useful. And why do we have node information in neighborhood table? It would be extremely repetitive. Is there any way we can use neighborhood table to only store edge information and then have another table to store node information at the same time as neighborhood table?

Thanks for the great comments! This PR wasn't meant to reflect the final state of protobuf data fields whatsoever. I planned to talk about them with both of you (@barkincavdaroglu  and @ajmcquilkin) and update the generator to reflect whatever we decide them to be.
  
As far as I understand it, we'll want to assemble an edge table with the following info (and possibly repeating edges):
```
| Node 1  |   Node 2  |  snr  |
| --------|----------|------|
       0              1             0.1   
       1              2             0.2
```
The question is how to get this information in the network. I'm assuming we want to signal to the nodes to broadcast protobufs with their own info to all their neighbors for a certain amount of time. This will inform both the way we get snr and the way we populate the neighbor lists (which is why those don't make sense right now).

As soon as I know that, I'll fix those problems.

uhuruhashimoto:
> Looks really solid! One thing I'm wondering is if it's worth maintaining the last time that each node has heard from each of its neighbors. I'm working on a design doc now for what this proto should look like, and one idea that came up was that we could include a logistic factor of time within the edge weight calculation, since if we haven't heard from a node in a while that often means the connection has dropped.

Included received time for each edge, and sent time for the packet as a whole (see `neighbor_info.rs`).

src-tauri/src/mocks/mock_generator.rs
uhuruhashimoto: Update: We are including the node id and a sent timestamp, and an id, received timestamp, and snr for each edge. The node ids will be used to get the position info that we already maintain (which includes latitude, longitude, speed, and heading). 

So that info will be available as long as we can retrieve it using node ids.

barkincavdaroglu:
Am I to understand that the example you gave produces a graph with 3 nodes where node 0 has undirected edges to node 1 and 2? Just double checking!

uhuruhashimoto:
> Am I to understand that the example you gave produces a graph with 3 nodes where node 0 has undirected edges to node 1 and 2? Just double checking!

Yes, in this case I ran `mock_generator` with a 3-node graph with 80% connectivity and got the following results:
```
Node 0
- recorded message from Node 1
Node 1
- recorded message from Node 0
Node 2
- recorded message from Node 0
- recorded message from Node 1
```

The connectivity refers to number of possible directed edges that are applied to the graph. In this case, we obtain a fully connected graph of undirected edges. However, not all our nodes register messages from all other nodes. If we drop the percentage further, we end up with truly disconnected nodes. 

The number of edges on our graph depends directly on how we construct it. This PR is meant to be constrained to the generator itself (generating the record above), so I'll address directed/undirected edge construction along with location mapping in a follow-up graph construction PR.

PR #190, opened 2023-01-12
Change notifications to work in app

### LINKED ISSUE #191: Notifications does not work on tauri app
On the tauri app, the notification "This functionality is not implemented" does not display when clicking the buttons on the map interaction pane.

sherry1iu:
image

This PR fixes a bug where the popup notification only works on the chrome app, and not the webapp. This PR changes the type of notification and standardizes it.

Closes #191

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,8 @@
     "maplibre-gl": "^2.4.0",
     "prop-types": "^15.8.1",
     "react": "^18.2.0",
+    "react-alert": "^7.0.3",
+    "react-alert-template-basic": "^1.0.2",
     "react-dom": "^18.2.0",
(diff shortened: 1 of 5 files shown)


ajmcquilkin:
Looks solid, but I would suggest putting an icebox ticket on the board for removing the two installed packages once we don't need the alert functionality.

ajmcquilkin (changes requested):
Functionality looks solid, but could you regenerate the `pnpm-lock.yaml` file? It looks like your formatter modified the lockfile, which we don't want due to it being autogenerated. Could you delete the file then run `pnpm i` again? That should reset the file

PR #195, opened 2023-01-13
Reads active node info, sets as default values for device config

### LINKED ISSUE #172: Set defaults in settings UI
Read in a node's current name, and have that set as the default in the settings UI instead of having a placeholder. 

This issue also doubles as a learning ticket for react redux.

sherry1iu:
PR adds in code that reads the active node's information to populate default imputs in device config. 

settings init

Also changed appearance of alerts:
image

Closes #172


src/components/Settings/Settings.tsx
```diff
@@ -84,6 +99,7 @@ const Settings = () => {
                 minLength={1}
                 value={deviceNickname}
                 onChange={(e) => setDeviceNickname(e.target.value)}
+                disabled={selectActiveNode === null ? true : false}
```
ajmcquilkin, line 102: Nit: you don't need a ternary here, the condition already evaluates to true or false

src/components/Settings/Settings.tsx
```diff
@@ -68,6 +82,7 @@ const Settings = () => {
                 label="Personalised name for this device."
                 value={deviceName}
                 onChange={(e) => setDeviceName(e.target.value)}
+                disabled={selectActiveNode == null ? true : false}
```
ajmcquilkin, line 85: Nit: same here

ajmcquilkin (approved):

src/components/Settings/Settings.tsx
sherry1iu: Added it because I got tired of seeing the red lines lol

ajmcquilkin: What red lines were you seeing? That should evaluate to a boolean expression

sherry1iu: Just realized I was evaluating the function instead of calling it; will make new commit to fix this

ajmcquilkin: Oh yeah, note that you shouldn't evaluate selectors inline. I would call it with the `useSelector` hook outside of the JSX then use that resulting value

ajmcquilkin: Looks like you could use the `activeNode` variable here

sherry1iu: That works, thanks!

src/components/Settings/Settings.tsx
```diff
@@ -84,6 +99,7 @@ const Settings = () => {
                 minLength={1}
                 value={deviceNickname}
                 onChange={(e) => setDeviceNickname(e.target.value)}
+                disabled={activeNode ? false : true}
```
ajmcquilkin, line 102: Nit: could also use `!activeNode` which converts to `true` only if `activeNode` is `null`

ajmcquilkin:
Should be good to merge! Lmk and I can approve once the PR is marked as ready for review.

sherry1iu:
I want to play with it a little more, but thanks!

ajmcquilkin (approved):

PR #196, opened 2023-01-15
Uhuruhashimoto/graph construction

uhuruhashimoto:
## Goal 
This PR adds support for a location hashmap of type `{u32, MeshNode}` to retrieve location info for nodes during graph construction.

## Implementation
- This PR adds a location creation function in the `mocks` module that creates a hashmap with randomly generated latitude, longitude, and altitude (and all other fields zeroed). 
- It also adds the hashmap into the `load_graph` function, which is used to initialize graphs. In that function, we take the SNR from the `NeighborInfo` packets, and ignore the SNR from the location hashmap (as it is assumed to be older). 
- It also cleans up the previous tauri event handlers used to communicate from the frontend to call algorithms, as these are now obsolete (and will be addressed from the design level up in the backend-to-front epic). 

Note: The weighting of contradictory SNRs is currently not done by timestamp. It is currently done within graph construction itself (Unsure if the graph creates two directed edges?). Edge weights do not currently have timestamps attached to them. 
Normalization of edge weights/SNRs can be added to this PR, or extended to a follow-up PR if we want to keep PR size down. 

## Testing
Unittests pass. None added.


barkincavdaroglu:
Our graph is undirected as it uses ```petgraph::stable_graph::StableUnGraph```. We might need to change a few stuff if we want to model the network as directed graph. 
I thought that if at time x b communicated with a then a would also be able to communicate with b at x. In this case the graph can remain undirected.

barkincavdaroglu (approved):

barkincavdaroglu:
Approved it because the discussion on undirected vs directed does not affect this PR. All the changes would be in graph struct itself.

uhuruhashimoto:
> Our graph is undirected as it uses `petgraph::stable_graph::StableUnGraph`. We might need to change a few stuff if we want to model the network as directed graph. I thought that if at time x b communicated with a then a would also be able to communicate with b. In this case the graph can remain undirected. Thoughts? @ajmcquilkin

@barkincavdaroglu what currently happens if you pass two edges `{A, B}` and `{B, A}` into `edge_factory`? Does our graph treat these as separate edges, or normalize them into a single edge? 
In the case above we'd have two different SNR values for each directed edge. It might be worth dealing with this in the initialization function because we might want to take the most recent value (for a single `{A,B}` undirected edge), and currently our timestamps don't persist past graph initialization.

barkincavdaroglu:
It would treat them as separate edges, though in petgraph graph the previously added would be replaced by the new one. Let's talk about this during our meeting today.

uhuruhashimoto:
> It would treat them as separate edges, though in petgraph graph the previously added would be replaced by the new one. Let's talk about this during our meeting today.

Sounds good. Merging and will follow up with another PR to address this.

PR #201, opened 2023-01-16
Data Storage

### LINKED ISSUE #147: Handle Disconnections During Timeline Construction
This issue will handle the case where network disconnects. The way we identify a disconnection is checking the number of nodes in consecutive snapshots. That is, if abs(|V_i| - |V_(i + 1)|) >= 1, then network is disconnected because there are some nodes unreachable from the central node (admin). 

When this happens, we partition the current timeline into 2: Timeline 1 is the series of snapshots until V_(i + 1) (which we label as "failed" for ML models), and Timeline 2 starts with when abs(|V_j| - |V_(j + 1)|) = 0 and ends when another disconnection occurs or the rescue attempt terminates. The portion (which may be the whole timeline) will be labeled as "successful" if no disconnection happens.

### LINKED ISSUE #200: Finalize Data Pipeline
Convert Postgre in #143 to local text files.

barkincavdaroglu:
Closes #200. Resolves #147 

This PR has a data structure Timeline that will be initialized for each rescue attempt to write to local data directory using txt files. For the purposes of MVP we keep a directory named data in src-tauri directory with subdirectory called timelines where we store snapshots of a timeline in its respective file. We also keep a simple text file data/labels.txt where each line is (timeline_id, label), where label is a binary value.

The Timeline contains the most current snapshot. Each time a new snapshot is taken, it checks whether a disconnection has occurred or not using the difference in number of nodes between the snapshot timeline contains and the snapshot that is passed into add_snapshot function (of Timeline). If there is a disconnection, we end the current timeline, label it as 0 in training table, and start a new timeline whose first snapshot is the function parameter.

The test expects data and data/timelines directories in src-tauri.

A sample timeline file with 2 snapshots has the following structure:
```txt
4
O: a 0 lat 43.71489 lon -72.28486 alt 1 v 0 th 0 deg 0.9
O: b 1 lat 43.71584 lon -72.28239 alt 1 v 0 th 0 deg 1.51
O: c 2 lat 43.7114 lon -72.28332 alt 1 v 0 th 0 deg 0.79
O: d 3 lat 43.71235 lon -72.28085 alt 1 v 0 th 0 deg 0.6
E: a b 0.51
E: a c 0.39
E: b c 0.4
E: b d 0.6

4
O: b 0 lat 43.71584 lon -72.28486 alt 1 v 0 th 0 deg 1.36
O: a 1 lat 43.71489 lon -72.28239 alt 1 v 0 th 0 deg 0.9299999999999999
O: d 2 lat 43.71235 lon -72.28085 alt 1 v 0 th 0 deg 0.65
O: c 3 lat 43.7114 lon -72.28332 alt 1 v 0 th 0 deg 0.44
E: a b 0.6
E: a c 0.33
E: b d 0.65
E: b c 0.11
```

And data/labels.txt has the following structure:
```txt
1,1
```
Since there was no disconnection in this timeline.


src-tauri/src/aux_data_structures/timeline.rs
```diff
@@ -0,0 +1,234 @@
...
+
+        let filename = format!("{}/{}.txt", self.data_dir, timeline_id);
+        let mut file = OpenOptions::new()
+            .append(true)
+            .create(true)
+            .open(filename)
+            .unwrap();
+
+        if let Err(e) = writeln!(file, "{}", snapshot_string) {
+            eprintln!("Couldn't write to file: {}", e);
```
uhuruhashimoto, line 107: Nitpick: Do we have a defined failure behavior if the user doesn't have these directories (printing a clear error message w instructions, or creating them)? If we don't create them, can we note that we need them in a docstring/readme/centralized place?

barkincavdaroglu: We can either define that location ourselves or leave it up to the user. If the user enters in a location only root user has access to, the app will probably crash though.

uhuruhashimoto: Sounds good! Mind adding some detail to the error messages/comments regarding that?

barkincavdaroglu: Best i can do is:
```rust
if let Err(e) = writeln!(labels_file, "{}", label_text) {
    eprintln!("Couldn't write label to file {}: {}", filename.clone(), e);
}
```
Print which file caused the error followed by the error message. We can't know what caused the error, though the error message should be somewhat straightforward.

uhuruhashimoto: oh, missed the error in that println. Looks good!

uhuruhashimoto:
Looks great Barkin! I don't have any high-level critiques of this PR, just a few open questions/ideas -- I'm assuming we're settled on txt files for MVP. If so, we should document this design choice somewhere (docstring in the `timeline.rs` file?). Also not sure if we should ask the user permission eventually? 

Those are broader questions though, and the timeline implementation of this PR looks good to me.

src-tauri/src/aux_data_structures/timeline.rs
```diff
@@ -0,0 +1,247 @@
...
+                }
+                true
+            }
+        }
+    }
+
+    pub fn get_curr_snapshot(&self) -> Option<Graph> {
+        match self.curr_snapshot {
+            None => None,
+            Some(ref curr_snapshot) => Some(curr_snapshot.clone()),
```
ajmcquilkin, line 97: Is there a reason we're duplicating the current snapshot? This would mean that it wouldn't be possible to edit the snapshot being received. This cloning might be something we leave to the consumer of this function

src-tauri/src/aux_functions/take_snapshot.rs
```diff
@@ -37,6 +39,7 @@ pub fn total_distance(lat1: f64, lon1: f64, alt1: f64, lat2: f64, lon2: f64, alt
 
 fn save_relative_ordering(graph: &Graph, graph_text: &mut String) {
     let mut nodes = graph.get_nodes().clone();
+    let node_fts: HashMap<String, String> = take_snapshot_of_node_fts(graph);
```
ajmcquilkin, line 42: Do we need this explicit typing here?

ajmcquilkin (commented):
Left some thoughts, but overall solid. I'm curious if we can give the user the option to specify where the files are being saved, since once this is bundled into an executable these files are going to be saved in some obscure system directory that the user won't necessarily know about.

src-tauri/Cargo.lock
```diff
@@ -78,7 +78,9 @@ dependencies = [
  "async-trait",
```
uhuruhashimoto, line 78: Unsure if we want the `lock` file, maybe change the `toml` file instead (though I don't see new packages inline)? Is this a merge leftover?

barkincavdaroglu: I did not manually change lock file. This branch carried over changes from the branch where I implemented postgresql, so toml file also included those extra packages. Since we don't use them anymore, I removed them from toml, which triggered the changes in lock (not visible in the last commit, but in earlier ones).

barkincavdaroglu:
> Looks great Barkin! I don't have any high-level critiques of this PR, just a few open questions/ideas -- I'm assuming we're settled on txt files for MVP. If so, we should document this design choice somewhere (docstring in the `timeline.rs` file?). Also not sure if we should ask the user permission eventually?
> 
> Those are broader questions though, and the timeline implementation of this PR looks good to me.
> 
> Edit: Holding off on approving for now because I'm not sure I fully understand the broader context of this PR and want more eyes on it :)

Yes we settled on txt files. 

> Looks great Barkin! I don't have any high-level critiques of this PR, just a few open questions/ideas -- I'm assuming we're settled on txt files for MVP. If so, we should document this design choice somewhere (docstring in the `timeline.rs` file?). Also not sure if we should ask the user permission eventually?
> 
> Those are broader questions though, and the timeline implementation of this PR looks good to me.
> 
> Edit: Holding off on approving for now because I'm not sure I fully understand how local storage in this PR will affect the product as a whole, and what our plan is for obtaining user permissions/giving the user control over the way we use their storage.

Timeline constructor takes in a dictionary of paths directing to where the data will be stored. The functionality to create directory and asking for user permission does not belong to Timeline object, so it's separate from this PR.

ajmcquilkin (approved):

uhuruhashimoto:
> > Looks great Barkin! I don't have any high-level critiques of this PR, just a few open questions/ideas -- I'm assuming we're settled on txt files for MVP. If so, we should document this design choice somewhere (docstring in the `timeline.rs` file?). Also not sure if we should ask the user permission eventually?
> > Those are broader questions though, and the timeline implementation of this PR looks good to me.
> > Edit: Holding off on approving for now because I'm not sure I fully understand how local storage in this PR will affect the product as a whole, and what our plan is for obtaining user permissions/giving the user control over the way we use their storage.
> 
> Yes we settled on txt files. Timeline constructor takes in a dictionary of paths directing to where the data will be stored. The functionality to create directory and asking for user permission does not belong to Timeline object, so it is meant to be separate from this PR. However, if you guys believe that it's necessary to address that with this PR, then we would need to discuss where to save the data. I would rather we discuss that soon.

Sounds good!

ajmcquilkin (approved):

PR #205, opened 2023-01-18
Added function to create adjacency matrix

### LINKED ISSUE #203: ADJ Matrix
Create aux_function that creates adjacency matrix given a graph. Part of #202.

barkincavdaroglu:
Closes #203.

This PR implements a simple function that takes in a graph and returns its adjacency matrix.

### DIFF
--- a/src-tauri/src/aux_functions/adj_matrix.rs
+++ b/src-tauri/src/aux_functions/adj_matrix.rs
@@ -1,0 +1,104 @@
+use crate::graph::graph_ds::Graph;
+use std::collections::HashMap;
+
+/// Converts a graph to an adjacency matrix.
+///
+/// # Arguments
+///
+/// * `graph` - a graph
+///
+/// # Returns
+///
+/// * `Vec<Vec<f64>>` - an adjacency matrix
+pub fn convert_to_adj_matrix(graph: &Graph) -> Vec<Vec<f64>> {
+    let mut adj_matrix: Vec<Vec<f64>> = Vec::new();
+    let nodes = graph.get_nodes();
+    let edges = graph.get_edges();
+
+    let mut ordering_counter = 0 as usize;
+    let mut node_id_to_int = HashMap::new();
+
+    for node in nodes.clone() {
+        node_id_to_int.insert(node.name.clone(), ordering_counter);
+        ordering_counter += 1;
+    }
+
+    for _ in 0..nodes.len() {
+        let mut row: Vec<f64> = Vec::new();
+        for _ in 0..nodes.len() {
+            row.push(0.0);
+        }
+        adj_matrix.push(row);
+    }
+
+    for edge in edges {
+        let u_name = graph.get_node(edge.get_u()).name.clone();
+        let v_name = graph.get_node(edge.get_v()).name.clone();
+
+        let u_id = node_id_to_int.get(&u_name).unwrap();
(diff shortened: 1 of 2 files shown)


barkincavdaroglu:
Don't delete branch after merge please.

src-tauri/src/aux_functions/adj_matrix.rs
```diff
@@ -0,0 +1,104 @@
...
+/// * `Vec<Vec<f64>>` - an adjacency matrix
+pub fn convert_to_adj_matrix(graph: &Graph) -> Vec<Vec<f64>> {
+    let mut adj_matrix: Vec<Vec<f64>> = Vec::new();
+    let nodes = graph.get_nodes();
+    let edges = graph.get_edges();
+
+    let mut ordering_counter = 0 as usize;
+    let mut node_id_to_int = HashMap::new();
+
+    for node in nodes.clone() {
```
ajmcquilkin, line 21: Why are we cloning this?

src-tauri/src/aux_functions/adj_matrix.rs
```diff
@@ -0,0 +1,104 @@
...
+        node_id_to_int.insert(node.name.clone(), ordering_counter);
+        ordering_counter += 1;
+    }
+
+    for _ in 0..nodes.len() {
+        let mut row: Vec<f64> = Vec::new();
```
ajmcquilkin, line 27: Don't need explicit typing here

src-tauri/src/aux_functions/adj_matrix.rs
```diff
@@ -0,0 +1,104 @@
...
+            row.push(0.0);
+        }
+        adj_matrix.push(row);
+    }
+
+    for edge in edges {
+        let u_name = graph.get_node(edge.get_u()).name.clone();
+        let v_name = graph.get_node(edge.get_v()).name.clone();
+
+        let u_id = node_id_to_int.get(&u_name).unwrap();
```
ajmcquilkin, line 38: Can we use `expect(&str mesg)` instead? In the event this code panics, `unwrap` won't tell the user the human-readable meaning of the problem

src-tauri/src/aux_functions/adj_matrix.rs
```diff
@@ -0,0 +1,104 @@
...
+    use crate::graph::{edge::Edge, node::Node};
+
+    use super::*;
+
+    #[test]
+    fn test_adj_matrix() {
+        let mut G1 = Graph::new();
+
+        // Create a few nodes and edges and add to graph
+        let a: String = "a".to_string();
```
ajmcquilkin, line 61: Nit: same as above, no explicit typing needed here

ajmcquilkin (approved):
Looks good, just some code style comments

src-tauri/src/aux_functions/adj_matrix.rs
barkincavdaroglu: Because we are borrowing nodes later on. But we can also use &nodes here instead of cloning. I fixed this and it will be included in my next pr.

src-tauri/src/aux_functions/adj_matrix.rs
barkincavdaroglu: This would panic if a node is part of an edge but not in the graph. In our add_edge we already give an error if node doesn't exist, so the program would never panic here.

PR #206, opened 2023-01-18
Added a function to calculate diffusion centralities

### LINKED ISSUE #202: Implement Diffusion Centrality Algorithm
https://www.nber.org/system/files/working_papers/w20422/w20422.pdf

Diffusion centrality measures how extensively the information spreads as a function of the initial node. The algorithm creates a "hearing matrix' so that the i-th row is the expected total number of times that some piece of information that originates from i is heard by any of the members of the graph during a T-period time interval.

barkincavdaroglu:
Closes #202 

This algorithm is a probabilistic simulation of network reliability in T time periods. Specifically, given node a, it calculates the expected number of times a will be able to communicate with the rest of the network. 

It returns a 2D map with the following format:
```txt
{
  “a”: {
	“a”: sum of its scores,
	“b”: E(ab),
	“c”: E(ac),
  },
  “b”: {
	“a”: E(ab),
	“b”: sum of its scores,
	“c”: E(bc),
  },
  “c”: {
	“a”: E(ac),
	“b”: E(bc),
	“c”: sum of its scores,
  }
}
```

The algorithm utilizes matrix powers to simulate paths between all possible nodes and multiplies it with probability q, which is estimated via the maximal eigenvalue v of adjacency matrix where average degree <= v <= max degree. If we consider that in each time unit interval a node tries to communicate with network, then parameter T allows us to simulate this for T time.

### DIFF
--- a/src-tauri/src/algorithms/diffcen.rs
+++ b/src-tauri/src/algorithms/diffcen.rs
@@ -4,6 +4,17 @@ use crate::aux_functions::adj_matrix::convert_to_adj_matrix;
 use crate::graph::graph_ds::Graph;
 use nalgebra::DMatrix;
 
+/// Calculates the diffusion centrality of each node in the graph.
+/// Diffusion centrality is a measure of how much information a node
+/// can diffuse to other nodes in the graph.
+///
+/// # Arguments
+///
+/// * `g` - The graph to calculate diffusion centrality for.
+///
+/// # Returns
+///
+/// * `Option<HashMap<String, HashMap<String, f64>>>` - A hashmap of node ids to a hashmap of node ids to diffusion centrality values.
 pub fn diffusion_centrality(g: &Graph) -> Option<HashMap<String, HashMap<String, f64>>> {
     let (adj_matrix, int_to_node_id) = convert_to_adj_matrix(g);
     let mut node_to_diffcen = HashMap::new();
--- a/src-tauri/src/algorithms/diffcen.rs
+++ b/src-tauri/src/algorithms/diffcen.rs
@@ -15,7 +15,7 @@ use nalgebra::DMatrix;
 /// # Returns
 ///
 /// * `Option<HashMap<String, HashMap<String, f64>>>` - A hashmap of node ids to a hashmap of node ids to diffusion centrality values.
-pub fn diffusion_centrality(g: &Graph) -> Option<HashMap<String, HashMap<String, f64>>> {
+pub fn diffusion_centrality(g: &Graph, T: u32) -> Option<HashMap<String, HashMap<String, f64>>> {
     let (adj_matrix, int_to_node_id) = convert_to_adj_matrix(g);
     let mut node_to_diffcen = HashMap::new();
 
@@ -50,7 +50,7 @@ pub fn diffusion_centrality(g: &Graph) -> Option<HashMap<String, HashMap<String,
 
             let mut H = DMatrix::zeros(g.get_nodes().len(), g.get_nodes().len());
 
-            for t in 1..6 {
+            for t in 1..T + 1 {
                 H += (q * &matrix).pow(t) * &identity_matrix;
             }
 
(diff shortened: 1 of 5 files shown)


src-tauri/src/algorithms/diffcen.rs
```diff
@@ -0,0 +1,112 @@
+use std::collections::HashMap;
+
+use crate::aux_functions::adj_matrix::convert_to_adj_matrix;
+use crate::graph::graph_ds::Graph;
+use nalgebra::DMatrix;
+
+/// Calculates the diffusion centrality of each node in the graph.
+/// Diffusion centrality is a measure of how much information a node
```
ajmcquilkin, line 8: How should the user interpret a `None` value coming out of this function?

src-tauri/src/aux_functions/adj_matrix.rs
```diff
@@ -10,21 +10,24 @@ use std::collections::HashMap;
 /// # Returns
 ///
 /// * `Vec<Vec<f64>>` - an adjacency matrix
-pub fn convert_to_adj_matrix(graph: &Graph) -> Vec<Vec<f64>> {
-    let mut adj_matrix: Vec<Vec<f64>> = Vec::new();
+pub fn convert_to_adj_matrix(graph: &Graph) -> (Vec<Vec<f64>>, HashMap<usize, String>) {
```
ajmcquilkin, line 13: I'm hesitant to return a tuple in Rust since it can be somewhat arbitrary as to the meaning of the result. I'd probably lean towards making this a new `struct` type

src-tauri/src/aux_functions/adj_matrix.rs
```diff
@@ -44,7 +47,7 @@ pub fn convert_to_adj_matrix(graph: &Graph) -> Vec<Vec<f64>> {
         adj_matrix[*v_id][*u_id] = weight;
     }
 
-    return adj_matrix;
+    return (adj_matrix, int_to_node_id);
```
ajmcquilkin, line 50: See comment above

ajmcquilkin (commented):
I can't speak to the correctness of the diffusion code, but just some comments. Would love thoughts from @uhuruhashimoto too on the `tuple` vs `struct`

src-tauri/src/algorithms/diffcen.rs
barkincavdaroglu: That will be interpreted by the global state I am working on. Basically it depends on whether the eigenvalues are real or not. None if complex result if real.

src-tauri/src/aux_functions/adj_matrix.rs
barkincavdaroglu: This will also be handled by global state. However, even then, I was thinking of returning tuple because this function won't be exposed to anything other than the global state struct, which will know how to handle the results of this function.

barkincavdaroglu: Also, that's the point of tuples right? They are ordered so you know which element is what.

src-tauri/src/algorithms/diffcen.rs
ajmcquilkin: Okay good to know. I ask the question more as a suggestion for the docs since I didn't see any context on the `Option` return type

src-tauri/src/aux_functions/adj_matrix.rs
ajmcquilkin: Yeah that is the point, but they provide no information on how the tuple is meant to be consumed. Seeing `t[0]` is way less informative then `t.matrix`, especially if that tuple grows to contain multiple fields of the same type

barkincavdaroglu: Well in this instance I think it's more efficient since the size will not grow and the elements are of different types. The function signature is clear enough that you'll know t[0] is adj matrix because adj matrix cannot be represented via HashMap<usize, String>. But again, nobody really needs to know this because state will have fields adj_matrix and etc.

ajmcquilkin (approved):

PR #210, opened 2023-01-21
Update protobuf genereation to handle proto namespacing change

### LINKED ISSUE #208: Bump Protos Version
Update protobufs to `2.0.14` to add support for waypoint emojis and namespacing (among other things).

ajmcquilkin:
This PR updates the build flow to handle namespacing changes introduced in [protobufs/#276](https://github.com/meshtastic/protobufs/pull/276). This PR also adds two additional `match` cases for handling `QueueStatus` and `XmodemPacket` data types.

Closes #208

### DIFF
--- a/src-tauri/build.rs
+++ b/src-tauri/build.rs
@@ -2,13 +2,13 @@ use prost_build;
 use walkdir::WalkDir;
 
 fn main() -> std::io::Result<()> {
-    println!("cargo:rerun-if-changed=protobufs");
+    let protobufs_dir = "protobufs/meshtastic/";
+    println!("cargo:rerun-if-changed={}", protobufs_dir);
 
     // Allows protobuf compilation without installing the `protoc` binary
     let protoc_path = protoc_bin_vendored::protoc_bin_path().unwrap();
     std::env::set_var("PROTOC", protoc_path);
 
-    let protobufs_dir = "protobufs";
     let mut protos = vec![];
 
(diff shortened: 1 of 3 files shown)


uhuruhashimoto:
Looks great! I have two questions before approval: 
1. In the match case for `handle_packet_from_radio` - where are we getting packet types from, and do we need to update it manually whenever we become aware of a new type of packet (ex. `XmodemPacket`, etc.)? Fine for MVP, just curious about the process for that. 
2. How was this PR tested?

ajmcquilkin:
1. Yes we will have to update each time a new packet comes out, although in theory we could also have a catch-all at the end that would remove the need for adding each packet. The reason I didn't use that is just a personal preference towards being more explicit about what packets we can handle.
2. The nice thing about this is that it updates the Rust build system, meaning that if it builds successfully then we should be good to go!

uhuruhashimoto:
> 1. Yes we will have to update each time a new packet comes out, although in theory we could also have a catch-all at the end that would remove the need for adding each packet. The reason I didn't use that is just a personal preference towards being more explicit about what packets we can handle.
> 2. The nice thing about this is that it updates the Rust build system, meaning that if it builds successfully then we should be good to go!

1. Agreed - the `match` is more readable, especially if we add comments explaining how we handle each type of packet.
2. Sounds good!

uhuruhashimoto (approved):

PR #211, opened 2023-01-21
Uhuruhashimoto/snr normalization

### LINKED ISSUE #198: Edge Normalization
Should graph edges be directed?
If not, how do we normalize graph edges with multiple SNR values for an edge? By timestamp? By average?

See discussion in https://github.com/ajmcquilkin/Meshtastic-emergency-response-client/pull/196

uhuruhashimoto:
## Goal
This PR updates graph initialization to create a single edge with the most recent SNR for any repeat edges.

## Implementation
We add the SNR and timestamp to a hashmap for each edge, and check it when creating an additional edge. We order the keys to prevent duplicate edges.
 
 ## Testing
 Added a unit test, `test_prioritize_earlier_snr`, that generates a 2-node graph with 1 edge containing the most recent SNR. 
 
 ## Notes / Questions About Parallel Edges
1. Do we want multiple parallel edges between nodes? If so, do we need this PR?
- Take a graph with nodes A and B. 
- If we have multiple directed edges between nodes ({A, B} and {B, A}), each edge SNR value would be a measurement of the signal strength from the other node. 
- In this example, our first edge is analogous to "A listening for signal from B" and the second to "B listening for signal from A". 
- If we have both edges, does it make sense to make them the same (most recent) value, which is what this PR does? 
 2. If we do want multiple edges, are we weighting those edges correctly when we initialize the graph?
 - The graph in question ends up with two directed edges, both correctly initialized using the most recent SNR, but they are weighted differently (although the distances should be identical). 
 3. Do we want to take the sum of parallel edges for the weight between nodes?
 - This gives us a value > 1 (which is not easily parsed to a SNR, which is what we were displaying on our designs).
 
 @barkincavdaroglu, any input? 
 
 ## Answers: 
 1. We currently support single edges; see conversation below. If we ever decide to abandon petgraph, it's easy to add parallel edges.
 2. We are now; see epsilon discussion below. May switch to softmax.
 3. Open question; to be addressed in future PR (this is a backend-to-frontend epic question)
 Closes #198


barkincavdaroglu:
I don't think we need parallel edges? None of the algorithms we have would be able to deal with such graph, and petgraph doesn't support it anyway (though our graph does). Also I think it might make sense to keep our graph undirected.

uhuruhashimoto:
> I don't think we need parallel edges? None of the algorithms we have would be able to deal with such graph, and petgraph doesn't support it anyway (though our graph does). Also I think it might make sense to keep our graph undirected.

Ok, in that case I'll update the graph init to only pass in a list of edges that are unique (no repeats), with the latest SNR value for each one. The graph itself will retain the ability to have multiple directed edges, but we won't give it any. The edges will be in [smaller_node_id, larger_node_id] directed order.

In `edge_factory`, we use the distance to normalize somehow -- do we still want to do that? How does that work?

barkincavdaroglu:
Wait our graph is undirected though.

barkincavdaroglu:
> In edge_factory, we use the distance to normalize somehow -- do we still want to do that? How does that work?

What do you mean?

uhuruhashimoto:
> Wait our graph is undirected though.

I think I'm misusing the term and I mean parallel edges. Our `Graph` (`StableGraph`) can have parallel {A, B} and {B, A} edges, even if the edges themselves are undirected. I mean parallel edges, not directed edges. 

My question is do we just want a single {A, B} edge with a single SNR value to display between nodes A and B?

barkincavdaroglu:
> I think I'm misusing the term and I mean parallel edges. Our Graph (StableGraph) can have parallel {A, B} and {B, A} edges, even if the edges themselves are undirected. I mean parallel edges, not directed edges.
> 
> My question is do we just want a single {A, B} edge with a single SNR value to display between nodes A and B?

StableGraph cannot handle parallel edges. When I say our graph does I mean the hashmap fields of graph_ds, which doesn't really mean anything because the actual graph (petgraph) replaces the old edge weight with the new weight when you try to add more than once edge between the same 2 nodes. So when there is an edge {A,B}, we would store the latest SNR.

src-tauri/src/aux_functions/graph_init.rs
```diff
@@ -53,7 +55,25 @@ pub fn load_graph(packets: Vec<NeighborInfo>, mut loc_hashmap: HashMap<u32, Mesh
...
+                } else {
+                    // Use the old snr
+                    radio_quality = alternate_snr;
+                }
+            } else {
+                // Because our node will be the next node's neighbor, add it to the hashmap in reverse order
+                snr_hashmap.insert((packet.id, neighbor.id), (neighbor.snr, neighbor.timestamp));
+            }
+            // Add edge to lists
             edge_left_endpoints.push(node_index);
```
barkincavdaroglu, line 77: I am just realizing this but this bit would create parallel edges since you are never checking if the edge already exists

barkincavdaroglu: It doesn't matter to functionality since our graph can't really handle parallel edges but edge_idx_map of graph would still keep track of these.

uhuruhashimoto:
> > I think I'm misusing the term and I mean parallel edges. Our Graph (StableGraph) can have parallel {A, B} and {B, A} edges, even if the edges themselves are undirected. I mean parallel edges, not directed edges.
> > My question is do we just want a single {A, B} edge with a single SNR value to display between nodes A and B?
> 
> StableGraph cannot handle parallel edges. When I say our graph does I mean the hashmap fields of graph_ds, which doesn't really mean anything because the actual graph (petgraph) replaces the old edge weight with the new weight when you try to add more than one edge between the same 2 nodes. So when there is an edge {A,B}, we would store the latest SNR. This is done automatically.
> 
> Parallel in this context is having multiple (either directed or undirected) edges between the same 2 nodes. When you use the notation {A,B} to denote an edge you are saying it's undirected, since that's a set notation and there is no ordering in sets. So {A,B} = {B,A}. Our graph is simple (no loops or parallel edges) and undirected so when we see an edge (B,A), a directed edge from B to A, we add both (B,A) and (A,B) with the same SNR to denote {A,B}. This is done automatically in add_edge function so you would just pass in (B,A).

Ohh gotcha, thanks! We should probably document that somewhere.

The reason I was confused is that the two edges generated in `StableGraph` had very different weights. This was due to an off-by-epsilon error in our `normalize_weight` function that would give us different values for the edge weight (A, B) and (B, A). I think this was related to the `fold`. 

Sample graph with different weights: 
```
StableGraph { Ty: "Undirected", node_count: 2, edge_count: 2, edges: (0, 1), (1, 0), node weights: {0: Node { name: "1", optimal_weighted_degree: 1.9444444444444444, longitude: 0.0, latitude: 0.0, altitude: 0.0 }, 1: Node { name: "2", optimal_weighted_degree: 1.9444444444444444, longitude: 0.0, latitude: 0.0, altitude: 0.0 }}, edge weights: {0: Edge { u: NodeIndex(0), v: NodeIndex(1), weight: 1.0 }, 1: Edge { u: NodeIndex(1), v: NodeIndex(0), weight: 0.9444444444444444 }}, free_node: NodeIndex(4294967295), free_edge: EdgeIndex(4294967295) }
```

I've changed the unit test to pass in representative distances, and changed the order of the addition in `normalize_weights` (which shouldn't have fixed anything), and now can't reproduce that bug. 

Current graph: 
```
StableGraph { Ty: "Undirected", node_count: 2, edge_count: 2, edges: (0, 1), (1, 0), node weights: {0: Node { name: "1", optimal_weighted_degree: 2.0, longitude: 0.0, latitude: 0.0, altitude: 0.0 }, 1: Node { name: "2", optimal_weighted_degree: 2.0, longitude: 0.0, latitude: 0.0, altitude: 0.0 }}, edge weights: {0: Edge { u: NodeIndex(0), v: NodeIndex(1), weight: 1.0 }, 1: Edge { u: NodeIndex(1), v: NodeIndex(0), weight: 1.0 }}, free_node: NodeIndex(4294967295), free_edge: EdgeIndex(4294967295) }
```
So we are now producing identically weighted edges in our `StableGraph` as intended!

src-tauri/src/aux_functions/graph_init.rs
uhuruhashimoto: it does, but should be fine. I could do the work to remove parallel edges here -- it might make things clearer. This whole function is also a monolithic mess, so I could clarify that. 

I left it because this module began as a wrapper that didn't make very many decisions, so I assumed that the graph creation itself would make decisions about which info to include/how to normalize info.

barkincavdaroglu:
> > > I think I'm misusing the term and I mean parallel edges. Our Graph (StableGraph) can have parallel {A, B} and {B, A} edges, even if the edges themselves are undirected. I mean parallel edges, not directed edges.
> > > My question is do we just want a single {A, B} edge with a single SNR value to display between nodes A and B?
> > 
> > 
> > StableGraph cannot handle parallel edges. When I say our graph does I mean the hashmap fields of graph_ds, which doesn't really mean anything because the actual graph (petgraph) replaces the old edge weight with the new weight when you try to add more than one edge between the same 2 nodes. So when there is an edge {A,B}, we would store the latest SNR. This is done automatically.
> > Parallel in this context is having multiple (either directed or undirected) edges between the same 2 nodes. When you use the notation {A,B} to denote an edge you are saying it's undirected, since that's a set notation and there is no ordering in sets. So {A,B} = {B,A}. Our graph is simple (no loops or parallel edges) and undirected so when we see an edge (B,A), a directed edge from B to A, we add both (B,A) and (A,B) with the same SNR to denote {A,B}. This is done automatically in add_edge function so you would just pass in (B,A).
> 
> Ohh gotcha, thanks! We should probably document that somewhere.
> 
> The reason I was confused is that the two edges generated in `StableGraph` had very different weights. This was due to an off-by-epsilon error in our `normalize_weight` function that would give us different values for the edge weight (A, B) and (B, A). I think this was related to the `fold`.
> 
> Sample graph with different weights:
> 
> ```
> StableGraph { Ty: "Undirected", node_count: 2, edge_count: 2, edges: (0, 1), (1, 0), node weights: {0: Node { name: "1", optimal_weighted_degree: 1.9444444444444444, longitude: 0.0, latitude: 0.0, altitude: 0.0 }, 1: Node { name: "2", optimal_weighted_degree: 1.9444444444444444, longitude: 0.0, latitude: 0.0, altitude: 0.0 }}, edge weights: {0: Edge { u: NodeIndex(0), v: NodeIndex(1), weight: 1.0 }, 1: Edge { u: NodeIndex(1), v: NodeIndex(0), weight: 0.9444444444444444 }}, free_node: NodeIndex(4294967295), free_edge: EdgeIndex(4294967295) }
> ```
> 
> I've changed the unit test to pass in representative distances, and changed the order of the addition in `normalize_weights` (which shouldn't have fixed anything), and now can't reproduce that bug.
> 
> Current graph:
> 
> ```
> StableGraph { Ty: "Undirected", node_count: 2, edge_count: 2, edges: (0, 1), (1, 0), node weights: {0: Node { name: "1", optimal_weighted_degree: 2.0, longitude: 0.0, latitude: 0.0, altitude: 0.0 }, 1: Node { name: "2", optimal_weighted_degree: 2.0, longitude: 0.0, latitude: 0.0, altitude: 0.0 }}, edge weights: {0: Edge { u: NodeIndex(0), v: NodeIndex(1), weight: 1.0 }, 1: Edge { u: NodeIndex(1), v: NodeIndex(0), weight: 1.0 }}, free_node: NodeIndex(4294967295), free_edge: EdgeIndex(4294967295) }
> ```
> 
> So we are now producing identically weighted edges in our `StableGraph` as intended!

Oh yeah I see why that might have been confusing. I added epsilon so that we would never get 0/0 but looks like I added it to the wrong variable lol. But that whole conversion will probably change because we (1) probably don't want to use minmax and (2) softmax makes more sense here.

src-tauri/src/aux_functions/graph_init.rs
barkincavdaroglu: So it doesn't cause any problems because as I said our algorithms and graph can't really handle parallel edges and these are automatically handled like you said. But the edge_idx map field would still store these parallel edges, and it's just a matter of memory overhead which is definitely not a significant problem because our graphs are really small, but it'd be great if you could take that into consideration just to be safe and future-proof!

barkincavdaroglu: Also I should note that the reason why I coded our graph in such a way that it would "technically" allow parallel edges is because in the future we might decide to make this a possibility and abandon petgraph.

uhuruhashimoto:
> 

Sounds good! I fixed `load_graph` to create only a single edge per node pair, so that version of the same-position edge case won't occur. However, if nodes don't get GPS info, they'll have their info initialized to 0, so we may hit a similar edge case. Should be good for this PR though.

src-tauri/src/aux_functions/graph_init.rs
uhuruhashimoto: Updated `load_graph` to only load a single edge for each pair of nodes. It should be an easy matter to change that to double edges if we ever want to support parallel edges with different weights. For now (and for simplicity's sake), we're loading a single edge.

src-tauri/src/aux_functions/graph_init.rs
```diff
@@ -16,50 +16,94 @@ To mock a graph, one can call the function load_graph() with a vector of Neighbo
...
 */
 
+pub fn add_node_to_graph_if_not_exists(graph: &mut Graph, node_id: u32) {
+    let name: String = node_id.to_string();
+    if !graph.contains_node(name.clone()) {
+        graph.add_node(name.clone());
+    }
+}
+
+pub fn as_key(node_1: u32, node_2: u32) -> (u32, u32) {
```
barkincavdaroglu, line 26: What's the purpose of this?

barkincavdaroglu (approved):

src-tauri/src/aux_functions/graph_init.rs
uhuruhashimoto: I'm using a hashmap to store edges in format [node, node] : [snr, timestamp] and didn't want to deal with collisions between node ids in the form of (A,B) and (B, A). That function standardizes the ordering of the node ids in the key.

PR #220, opened 2023-01-23
Uhuruhashimoto/organize mock and init

uhuruhashimoto:
## Goal 
Refactor the confusing `mock` utilities and integrate them in preparation for mocking a global hashmap of edges.

## Implementation
This PR renames the existing `mocks` module to `constructors`. It contains the following modules: 
- The `init` module contains creation utilities for `graph`, `edge_map`, and `node_map`
   - This creates data types with provided lists of packets.
- The `mock` module contains functions to generate random `MeshPackets`, and `NeighborInfoPackets`.
   - This creates random packet lists.
 
This PR also contains the start of a skeleton for mocking the edge map: 
- `mock_edge_map` stub in `mocks` (which we'll use location info to generate, since we don't have NeighborInfo packets).
- added `edges` hashmap to `MeshDevice` struct

## Testing
Unit test updated and created some new ones. All pass, except for the timing one (unrelated).


src-tauri/src/mocks/mocks.rs
```diff
@@ -120,20 +112,34 @@ pub fn generate_loc_info(num_nodes: i32) -> HashMap<u32, MeshNode> {
...
             data: node_info,
         };
-        loc_hashmap.insert(node_id as u32, meshnode);
+        meshnode_vec.push(meshnode);
     }
-    loc_hashmap
+    meshnode_vec
+}
+
+pub fn mock_edge_map_from_loc_info(
```
uhuruhashimoto, line 120: This is a stub we'll call from the handlers. To mock, we can pretend that position packets are neighborinfo packets. Every time we get one, we'll use our existing location info (ids and snrs) to generate a graph with random edges.

src-tauri/src/mesh/device/mod.rs
```diff
@@ -143,7 +143,8 @@ pub struct MeshDevice {
     pub config: protobufs::LocalConfig, // local-only device configuration
     pub my_node_info: protobufs::MyNodeInfo, // debug information specific to device
     pub nodes: HashMap<u32, MeshNode>, // network devices this device has communicated with
-    pub region_unset: bool,       // flag for whether device has an unset LoRa region
+    pub edges: HashMap<(u32, u32), (f64, u64)>, // edges between nodes (node_id, neighbor_id) -> (snr, timestamp)
```
uhuruhashimoto, line 146: Storing edges as a hashmap in the `MeshDevice` struct. While eventually this struct may contain raw `NeighborInfo` packets, they don't exist as of now, so we'll just store raw ids, snrs, and timestamps.

barkincavdaroglu:
There doesn't seem to be any comments explaining what each function does.

barkincavdaroglu:
Also now might be a good time to start adding markdown files in each subdirectory to explain the code. As we are now in the phase where there are interactions between subdirectories, we all should be able to go back to this markdown file to understand what's going on. This is not a blocker for this PR but we should probably create issues for documentation, what do you think @ajmcquilkin

src-tauri/src/mocks/init/init_edge_map.rs
```diff
@@ -0,0 +1,123 @@
+use crate::aux_data_structures::neighbor_info::{Neighbor, NeighborInfo};
+use std::collections::HashMap;
+
+pub fn init_edge_map(packets: Vec<NeighborInfo>) -> HashMap<(u32, u32), (f64, u64)> {
```
barkincavdaroglu, line 4: Since this is in src/mocks does it mean there will be another function that creates edges from real data? In other words, do you intend to use this function only for fake data? If not, why is it in this directory?

uhuruhashimoto:
> There doesn't seem to be any comments explaining what each function does.

Added docstring comments for all functions

src-tauri/src/mocks/init/init_edge_map.rs
uhuruhashimoto: You're right, that's a strange directory placement. Would it make more sense to change `mocks` module to a parent module of `constructors`, with submodules `mock` and `init`?

I had `init` functions for both maps in case we wanted to generate mock packets, populate hashmaps, and use those to create a graph. But since it makes more sense to add edges to our existing nodes at runtime, they're a little superfluous for the moment.

barkincavdaroglu: > Would it make more sense to change `mocks` module to a parent module of `constructors`, with submodules `mock` and `init`?
This makes much more sense I think.

uhuruhashimoto: Changed. Current file structure is: 
`constructors`
   - `init`
      - contains init functions for graph and edge and node hashmaps. Intended to allow us to run a full run of algorithms from the packets up
  - `mocks`
     - contains functions for generating random lists of `NeighborInfo` and `MeshNode` packets
     - contains stub for creating random edges from an existing hashmap of `MeshNode` structs (our location info)

barkincavdaroglu (approved):

Issue #166, opened 2023-01-09
Serial Disconnect and Reconnect Crashes Application


ajmcquilkin:
Currently, a user cannot cleanly disconnect from a serial device then reconnect to the same device without a `tauri::async_runtime::thread` error being thrown.

PR #222, opened 2023-01-24
Connect device user config to settings UI

### LINKED ISSUE #164: Connect settings modal to Rust backend
Allow for meaningful editing of settings using Tauri commands in the backend.

sherry1iu:
PR that connects rust backend to current version of settings UI. Allows for changing the `longname` and `shortname` of an active node connected through serial, according to user input in the settings pane. App restarts upon saving.

Will likely be refactored upon implementation of updated UI.
implement_changes

Closes #164

### DIFF
--- a/src/components/Settings/Settings.tsx
+++ b/src/components/Settings/Settings.tsx
@@ -6,9 +6,7 @@ import { XMarkIcon } from "@heroicons/react/24/solid";
 import { Input } from "@material-tailwind/react";
 
 import { selectActiveNode } from "@features/device/deviceSelectors";
-
 import type { User } from "@bindings/protobufs/User";
-
 import { invoke } from "@tauri-apps/api/tauri";
 
 // Function to convert decimal MAC address to hex
--- a/src/components/Settings/Settings.tsx
+++ b/src/components/Settings/Settings.tsx
@@ -29,37 +29,29 @@ const Settings = () => {
 
   // Functions to set device info. Forbidden non-null suppressed, but I checked if a device is connected.
   const [deviceID, setDeviceID] = useState(
-    activeNode ? activeNode.data.user!.id : "No device selected"
+    activeNode?.data.user?.id ?? "No device selected"
   );
   const [deviceName, setDeviceName] = useState(
-    activeNode ? activeNode.data.user!.longName : "No device selected"
+    activeNode?.data.user?.longName ?? "No device selected"
   );
   const [deviceNickname, setDeviceNickname] = useState(
-    activeNode ? activeNode.data.user!.shortName : "No device selected"
+    activeNode?.data.user?.shortName ?? "No device selected"
   );
 
   // MAC, Hardware, and Licensing. Even though they are not displayed when null, the check prevents an error on assignment
   const mac = activeNode
     ? convertMacAddr(activeNode.data.user!.macaddr)
     : "No device selected";
-  const hwModel = activeNode
-    ? activeNode.data.user!.hwModel
-    : "No device selected";
-  const isLicensed = activeNode
-    ? activeNode.data.user!.isLicensed
-    : "No device selected";
+  const hwModel = activeNode?.data.user?.hwModel ?? "No device selected";
+  const isLicensed = activeNode?.data.user?.isLicensed ?? "No device selected";
 
   // Submits the form. Triggered by pressing the save button
   const handleSubmit: FormEventHandler = (e) => {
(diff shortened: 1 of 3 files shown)


src/components/Settings/Settings.tsx
```diff
@@ -49,7 +51,23 @@ const Settings = () => {
...
-    e.preventDefault();
+    // e.preventDefault();
+    if (activeNode) {
+      const updatedUser: User = {
+        id: deviceID,
+        longName: deviceName,
+        shortName: deviceNickname,
+        macaddr: activeNode ? activeNode.data.user!.macaddr : [],
+        hwModel: activeNode ? activeNode.data.user!.hwModel : 0,
+        isLicensed: activeNode ? activeNode.data.user!.isLicensed : true,
```
uhuruhashimoto, line 62: Should this default to true without an active node?

src/components/Settings/Settings.tsx
```diff
@@ -49,7 +51,23 @@ const Settings = () => {
...
+      };
+
+      invoke("update_device_user", { user: updatedUser }).catch((e) => {
```
uhuruhashimoto, line 65: Commenting out of curiosity -- I know this has to be a string, but is there a way to have it not hardcoded when calling the event handler from the frontend? If not, do we want a standardized way of storing strings in the frontend (e.g. some data structure)? 

It seems like another level of abstraction and unnecessary for this PR, but I'm wondering if sometime in the future it would be nice to have a file with all the handler name strings and a description of what each of them does (and we'll invoke them using that). 

@sherry1iu and @ajmcquilkin - might be unnecessary, thoughts? I don't fully understand event handlers, so feel free to shoot this down

src/components/Settings/Settings.tsx
```diff
@@ -75,7 +93,7 @@ const Settings = () => {
           </div>
         </div>
         <div className="text-sm flex justify-center">
-          Note: This is just the UI, and is not currently connected to backend
+          Note: App will be reloaded on save
```
uhuruhashimoto, line 96: Mind explaining the thought behind this - 
- What do you mean by reloading?  
- Do we reload if we change the name info of our node, or any node? 
- Will we display the splash page whenever we update a node?

uhuruhashimoto:
Nice PR @sherry1iu! Left you some comments --  they're mostly about behavior, not style. Feel free to respond in kind and resolve. My TS is a little rusty (pardon the pun), so I'll wait to approve until someone on the frontend has looked it over. :)

src/components/Settings/Settings.tsx
ajmcquilkin: This legally cannot default to true, blocking PR on this. I wouldn't recommend completely rebuilding the `User` object, I would get the current device if it exists (if not I wouldn't even open the popup), then pull the user from that device's corresponding node. Lmk if you wanna chat on this

src/components/Settings/Settings.tsx
ajmcquilkin: Oh yeah this is probably a bigger issue that warrants its own ticket. The way we typically are handling these command calls is by creating a new redux saga handler that can then abstract over the call so that the string is only in one location. Not blocking this PR on this point, but I would encourage refactoring this into a similar form to the way we send string messages ([link](https://github.com/ajmcquilkin/Meshtastic-emergency-response-client/blob/21861dacb5d654b04945b8ddf0f6236ad310a61f/src/features/device/deviceSagas.ts#L65)). Also happy to chat on this one

src/components/Settings/Settings.tsx
ajmcquilkin: Maybe this is something I missed in a previous PR, but how does the app reload? And for the reload to be useful, we have to be able to guarantee that it takes longer than the device restart. I think the web client handles this with a "reconnect" command

ajmcquilkin (changes requested):
Blocking on the `isLicensed` issue

src/components/Settings/Settings.tsx
sherry1iu: Removed the ternary - the default would never have been used, and I only put it there because I thought it would throw a type error. Will  look into pulling the current `User`.

src/components/Settings/Settings.tsx
ajmcquilkin: Known bug with #166, no changes on this PR

ajmcquilkin (approved):
Looks great!

PR #223, opened 2023-01-24
Global State

barkincavdaroglu:
## **Analytics**
Module for interacting with the network / graph.

### **State**

The entry-point of this module is `State` struct in [state.rs](./state.rs). `State` accepts two parameters: `config_fields: HashMap<&str, &str>` and `is_save: bool` both of which are passed down to [Timeline](./aux_data_structures/timeline.rs). Former is a hashmap with two keys "timeline_data_dir" and "timeline_label_dir" representing the directory locations in which the data will be saved if `is_save` is true.

The overall structure of `State` is as below:

```rust
pub struct State {
    timeline: Timeline,
    history: History,
    time: SystemTime,
    algo_configs: AlgoConfig,
    algo_run_mode_auto: bool,
    algo_store: AlgoStore,
    algo_controller: AlgoController,
};
```

All fields of `State` are initialized as empty. When the backend centralizes the network information and creates a `Graph` object, it passes it to `State` as:

```rust
state.add_graph(&graph);
```

which the `state` adds to `state.timeline`.

Since the choice of algorithm to run is left to the user, frontend will send a bitfield of type `u8` to `state` to activate algorithms as:

```rust
state.set_algos(algos_bitfield: u8);
```

`algos_bitfield` can be constructed in the following way:

- To run the articulation point algorithm, the bitfield is 0b00001.
- To run the mincut algorithm, the bitfield is 0b00010.
- To run the diffusion centrality algorithm, the bitfield is 0b00100.
- To run the most similar timeline algorithm, the bitfield is 0b01000.
- To run the predicted state algorithm, the bitfield is 0b10000.
- To run all algorithms, the bitfield is 0b11111.

All other combinations can be derived from above.

Running **'activated'** algorithms is as simple as calling:

```rust
state.run_algos();
```

In this function, `State` gets the current snapshot from `Timeline` and calls `AlgoController` as:

```rust
self.algo_controller.run_algos(
    &curr_graph,
    &self.algo_configs,
    &mut self.history,
    &mut self.algo_store,
);
```

which itself calls the activated algorithms, stores their results as `Enum` in `AlgoStore` and logs the time in `History` object for each of the algorithms.

To retrieve algorithm results:

```rust
let algo_results = state.get_algo_results();
let ap_algo_res = algo_results.get_aps();
match ap_algo_res {
    APResult::Success(aps) => {
        println!("AP algorithm returned: {:?}", aps);
        assert_eq!(aps.len(), 0);
    }
    APResult::Error(err_str) => {
        panic!("Error running AP algorithm: {}", err_str);
    }
    APResult::Empty(b) => {
        panic!("AP algorithm returned empty result: {}", b);
    }
}
```
&nbsp;

### **AlgoStore**

This `Struct` saves the result of activated algorithms represented by their respective `Enum`s. It should be noted that the `AlgoStore` saves the latest results and not all of them. So if an algorithm is run 5 times, `AlgoStore` will only save the last one.

&nbsp;

### **AlgoController**

This `Struct` doesn't have any field but contains a group of functions that run algorithms and save and log their results. Only `State` interacts with `AlgoController` and no other module should interact with it.

&nbsp;

### **History**

For each algorithm `History` stores the time and elapsed time since prior time each algorithm was run.

&nbsp;

### **Enums**

For each algorithm, we implement an `Enum` that can take on any one of the three values: Success: Any, Error: String, Empty: bool. For example, for articulation point algorithm result, we have the following `Enum`:

```rust
#[derive(Debug)]
pub enum APResult {
    Success(Vec<NodeIndex>),
    Error(String),
    Empty(bool),
}
```

The type of `APResult::Success` is `Vec<NodeIndex>` because that's what articulation_point algorithm returns if there is no problem. The same concept applies to all other algorithm results.


src-tauri/src/global/algo_store.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    pub fn run_mincut(&mut self, g: &Graph) {
+        let sw_graph = &mut StoerWagnerGraph::new(g.clone());
+        let _mincut = stoer_wagner(sw_graph).unwrap();
+        let mut nodes_string = Vec::new();
+        for node in g.get_nodes() {
+            nodes_string.push(node.name.clone());
+        }
+        let mincut_edges = recover_mincut(sw_graph, nodes_string);
+        self.set_mincut(mincut_edges);
```
barkincavdaroglu, line 46: Only saving edges between mincuts here. Should we also save the weight?

src-tauri/src/global/algo_store.rs
```diff
@@ -0,0 +1,107 @@
...
+
+pub struct AlgoStore {
+    pub aps: Option<Vec<petgraph::graph::NodeIndex>>,
+    pub mincut: Option<Vec<Edge>>,
+    pub diff_cent: Option<HashMap<String, HashMap<String, f64>>>,
+    pub most_sim_t: Option<Graph>,
+    pub pred_state: Option<Graph>,
+}
+
+impl AlgoStore {
```
barkincavdaroglu, line 22: Does it make sense to store algo results here? Not sure. We can also save it in global state or create another struct.

barkincavdaroglu:
Going to move adj_matrix, d_adj_matrix, eigens, int_to_node_id and node_to_int to graph as optional fiels. It doesn't semantically make sense for GlobalState to hold these information as they are specific to graph snapshots.

uhuruhashimoto:
Also, this should likely be broken out into 2 PRs - (1) to add backend algo controls/state, and (2) to refactor existing modules. So the above feedback might be jumping ahead a bit

uhuruhashimoto:
One more question: on a high level, how do you envision this interacting with our existing `MeshDevice` struct that stores global device state? We'll pull nodes and edges from that to construct our Graph (via `constructors`, but otherwise they'll be separate places to store global state, right?

On a related note, how much of this is voluntary? That is, if the user does not opt in to storing data, how much of the global algo state will exist?

src-tauri/src/global/algo_store.rs
uhuruhashimoto: What does the weight represent? Is it an arbitrary number within the mincut problem (i.e. we're taking the max/min weighted edges), or on a set scale? (I don't quite understand mincut as a whole)

barkincavdaroglu:
> One more question: on a high level, how do you envision this interacting with our existing `MeshDevice` struct that stores global device state? We'll pull nodes and edges from that to construct our Graph (via `constructors`, but otherwise they'll be separate places to store global state, right?
> 
> On a related note, how much of this is voluntary? That is, if the user does not opt in to storing data, how much of the global algo state will exist?

It only interacts with the graph. So once the network is centralized, it would be passed into global state, which would then add it to timeline. Whether the user wants to save or not has no effect on this module, because we can still run algorithms without saving data and display the network on the map. I envision this as middle-person between the backend (mesh stuff) and frontend. Mesh backend does its stuff, centralizes the graph, and sends it to global state to be processed and used according to what the user demands (algorithms, displaying the graph, etc). History is only for frontend to display, in case user has picked manual mode for running algorithms, the time each algo was run. If we take snapshots every 5 minutes and the user hasn't run a certain algorithm for 12 minutes (as an example), then we can display a message on frontend to inform the user that the results are no longer applicable.

src-tauri/src/global/algo_store.rs
barkincavdaroglu: Weight of a mincut is the sum of weights of edges between the two cuts. Basically this algorithm is dividing set of nodes into two disjoint subsets such that the sum of weights of edges between these two subsets is minimal (no other such "cut" exists with smaller weight).

barkincavdaroglu:
> Have we added any new unit tests to cover the result enums?
> Mind adding a `Testing` section to the PR description to document how this PR was tested, given how big it is?

Not sure what you mean by unit tests to cover the result enums. I don't think these are necessary since their validity is already being tested by algorithms.

barkincavdaroglu:
> Initial thoughts:
> 
> * I like the algo store/controller/config/global division.
> * I like the error enums. Both are very clear to understand.
> 
> Can we have a top-level `algorithms` module, which contains the `aux` modules, and the module `global` (maybe renamed to `global_state`)? Right now our division of responsibility isn't very clear within the top-level backend modules.
> 
> Can we rename `state_err_enums` to `algo_results`?
> 
> Something like this:
> 
> * `algorithms`
>   
>   * `algorithm_functions`
>     
>     * [existing `algorithms` module]
>   * `algorithm_helpers`
>     
>     * [existing `aux_functions` module]
>     * [existing `aux_data_structures` module]
>   * `algorithm_results`
>     
>     * [all state_err_enums classes]
>   * `global_state` (maybe renamed to `algorithm_controls`)
>     
>     * [algo store/controller/config/global]

I don't think global state should be under algorithms because it uses timeline struct and is fundamentally a "level above" the algorithms.

src-tauri/src/global/algo_store.rs
uhuruhashimoto: Because the edges are weighted according to their SNR and distance, their weights are already difficult to understand (i.e. as a user looking at a number on the frontend). Because the weight of a mincut is the sum of these weights, I don't imagine a user will get anything from it unless we add a legend. 

So I'm gonna vote against saving/displaying the weight, if the main info we're trying to give the user is which edges make the graph most vulnerable to disconnection.

barkincavdaroglu: Fair enough, I agree. We won't have more than a few mincut edges anyway so the user can eyeball the sum if they want to.

uhuruhashimoto:
> > Have we added any new unit tests to cover the result enums?
> > Mind adding a `Testing` section to the PR description to document how this PR was tested, given how big it is?
> 
> Not sure what you mean by unit tests to cover the result enums. I don't think these are necessary since their validity is already being tested by algorithms. It would be akin to testing a struct without any functionality.

Oh my bad, I wasn't very clear. I didn't mean testing the enums/structs in isolation necessarily, was more curious if there was a way to test the algo state "middleman" incrementally (maybe via unit tests) before we try to pull algorithm results from the frontend.

barkincavdaroglu:
> Oh my bad, I wasn't very clear. I didn't mean testing the enums/structs in isolation necessarily, was more curious if there was a way to test the algo state "middleman" incrementally (maybe via unit tests) before we try to pull algorithm results from the frontend.

Not sure I quite understand what you mean. The algorithms already use these enums (+ they pass the unit tests implemented for algos), and these enums are saved in store -> they can be pulled from the frontend.

uhuruhashimoto:
> > Initial thoughts:
> > 
> > * I like the algo store/controller/config/global division.
> > * I like the error enums. Both are very clear to understand.
> > 
> > Can we have a top-level `algorithms` module, which contains the `aux` modules, and the module `global` (maybe renamed to `global_state`)? Right now our division of responsibility isn't very clear within the top-level backend modules.
> > Can we rename `state_err_enums` to `algo_results`?
> > Something like this:
> > 
> > * `algorithms`
> >   
> >   * `algorithm_functions`
> >     
> >     * [existing `algorithms` module]
> >   * `algorithm_helpers`
> >     
> >     * [existing `aux_functions` module]
> >     * [existing `aux_data_structures` module]
> >   * `algorithm_results`
> >     
> >     * [all state_err_enums classes]
> >   * `global_state` (maybe renamed to `algorithm_controls`)
> >     
> >     * [algo store/controller/config/global]
> 
> I don't think global state should be under algorithms because it uses timeline struct and is fundamentally a "level above" the algorithms.

I put the existing `algorithms` module under `algorithm_functions` -- they're 2 separate things. 
- `algorithms` (new upper level module)
   - `algorithm_functions` (previously `algorithms`)
   - `global_state`
   - `algorithm_results` (previously `state_err_enums`)
   
In my head, `algorithms` was basically a black box that contained all the info about (1) how we run specific algorithms, and (2) a controller that controls how we interact with the device backend/frontend.

So maybe this? 

- `analytics` (new upper level module)
  - `controller`
     - (everything in this PR)
  - `algorithms`
     - (algos and aux functions)

uhuruhashimoto (commented):
This is looking pretty clean. Module names make sense to me, and the idea of controller is well abstracted. 

Regression-wise, also not seeing any problems:
- I checked out the branch and can run all unit tests successfully.
- I can run the client itself without any problems. 

I'm trying to find a problem somewhere in the 1500 lines (!!), but not seeing any. 
Because this is such a big PR, I'll wait for @ajmcquilkin to get eyes on it before approving.

ajmcquilkin (approved):
Approved on behalf of @uhuruhashimoto

uhuruhashimoto:
> Approved on behalf of @uhuruhashimoto

Approved by proxy after pushing a commit and preventing myself from approving. Thanks @ajmcquilkin

Issue #94, opened 2022-11-05
Most Similar Graph Embeddings


barkincavdaroglu:
This is both a standalone module and also a helper for [#90](https://app.zenhub.com/workspaces/meshtastic---cosc-098-633f046aebd6cde319822187/issues/ajmcquilkin/meshtastic-emergency-response-client/90).

Let's say we take a snapshot of the network every 2 minutes, and that our average rescue attempt takes 2 hours. We will have 60 network snapshots per rescue attempt. If we save graph embeddings of these snapshots, then:

At time t, let's say we have a network of size n (# of nodes). Let's loop over the timelines (previous rescue attemps) with size n (so past rescue attempts with n nodes):

For some snapshot time k, compare the similarity (cosine or euclidean) between snapshot at t (our current snapshot) and snapshots at k and k - 2. We keep track of snapshot with best similarity and return it. The reason why we compare snapshots at both k and k - 2 is because we want a sequence most similar to ours and not just a point in time in that sequence.

After we find that sequence, we can use it to predict what the next state of our network at t + 1 will be. This is a standalone module because we can't train GNN model without enough data, and that most similar previous timelines may be sufficient to make a prediction. This is a helper module because even if we don't have enough data to train the GNN model at the beginning of rescue operation, we can find most similar past timelines to train the model.

One thing to note is that we can't just arbitrarily use cosine or euclidean on graph embeddings. We need to match the vertices somehow. We do the following:

Whenever we save an embedding of a snapshot, we label the vertices starting from left to right according to the haversine distance of each pair of nodes. Leftmost vertex is labeled 0. Closest vertex in terms of haversine distance is labeled 1. And so on. 

![Screenshot 2022-11-04 at 21.38.16.png](https://images.zenhubusercontent.com/5dd652f511a8290001d10cde/e1fb8f3a-8ae8-4797-af5d-10d953f45059)

Hence we get the relative positioning of vertices, which could then be used to match vertices of current snapshot.

Pointers on graph embeddings:

- https://github.com/benedekrozemberczki/graph2vec
- https://github.com/aditya-grover/node2vec
- https://karateclub.readthedocs.io/en/latest/modules/root.html#whole-graph-embedding

Haversine is described [here](https://en.wikipedia.org/wiki/Haversine_formula).

barkincavdaroglu:
[...]

PR #225, opened 2023-01-25
Uhuruhashimoto/distance conversion

uhuruhashimoto:
## Purpose
Centralize GPS conversion functions into a module/toolbox.

## Implementation
Distance conversion utilities are spread between `aux_functions` and `constructors`; make an upper-level data conversion module that both can pull from. Module only exposes *conversion* functions; it does not generate random data or expose conversion constants (`distance_constants`).

Note: also add a `gps floating point` -> `proto field` function that mocks can call.

## Tests
Unit tests updated and pass.

### DIFF
--- a/src-tauri/src/aux_functions/mod.rs
+++ b/src-tauri/src/aux_functions/mod.rs
@@ -1,5 +1,4 @@
 pub(crate) mod adj_matrix;
-pub(crate) mod conversion_factors;
 pub(crate) mod convert_to_graph_from_string;
 pub(crate) mod edge_factory;
 pub(crate) mod get_most_similar;
--- a/src-tauri/src/aux_functions/take_snapshot.rs
+++ b/src-tauri/src/aux_functions/take_snapshot.rs
@@ -1,41 +1,7 @@
 use std::{collections::HashMap, hash::Hash};
 
+use crate::distance_conversion::get_distance;
 use crate::graph::graph_ds::Graph;
-
-/// Returns Haversine distance between 2 nodes using their lat and long
-/// https://en.wikipedia.org/wiki/Haversine_formula
-///
-/// # Arguments
-///
-/// * `lat1` - latitude of node 1
-/// * `lon1` - longitude of node 1
-/// * `lat2` - latitude of node 2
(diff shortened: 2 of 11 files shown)


barkincavdaroglu (commented):
Since the conversion will only be used during centralization of graph, doesn't it make more sense to put this in init (or whatever directory will be responsible for constructing graph from messages)?

uhuruhashimoto:
> Since the conversion will only be used during centralization of graph, doesn't it make more sense to put this in init (or whatever directory will be responsible for constructing graph from messages)?

TLDR: Yes, except for a single little wrench in the works: `take_snapshot`. 

Currently, both `constructor` and `aux` modules use distance conversion: 
- `take_snapshot` uses the `total_distance` function
- `init_graph` uses the `get_distance` wrapper for the same function

This happens because we want to use the Haversine distance to (1) sort nodes in a snapshot, and (2) use the distance between nodes to add edges when we mock edges for our existing map of node locations.

My solution was to make a top-level `data_conversion` module that contains *only converters*, and that can be used by both modules.

@barkincavdaroglu  Mind explaining what `take_snapshot` does and why it needs to rank nodes by distance?

barkincavdaroglu:
> > Since the conversion will only be used during centralization of graph, doesn't it make more sense to put this in init (or whatever directory will be responsible for constructing graph from messages)?
> 
> TLDR: Yes, except for `take_snapshot`, which I don't understand enough to mess with.
> 
> Currently, both `constructor` and `aux` modules use distance conversion:
> 
> * `take_snapshot` uses the `total_distance` function
> * `init_graph` uses the `get_distance` wrapper for the same function
> 
> This happens because we want to use the Haversine distance to (1) sort nodes in a snapshot, and (2) use the distance between nodes to add edges when we mock edges for our existing map of node locations.
> 
> My solution was to make a top-level `data_conversion` module that contains _only converters_, and that can be used by both modules.
> 
> @barkincavdaroglu Mind explaining what `take_snapshot` does and why it needs to rank nodes by distance?

#97 and #94

barkincavdaroglu (approved):

Issue #230, opened 2023-01-26
Add request loading management


ajmcquilkin:
Currently we don't have a way to manage whether a given saga is in progress or whether it has succeeded or failed. We will need this to allow for our onboarding flow.

PR #231, opened 2023-01-26
Creating and styling onboarding page for client.

samfydega:
In this PR, I created and styled the page that will be used to onboard response coordinators to the emergency client. The main function of this page is to have the coordinator set-up their serial connection with their node. This ticket handles the design of the page. The next ticket will implement the functionality for setting up the serial connection with the back-end.

**Goal:**  Implement UI mockups for the client onboarding page.
**Implementation details:** Followed styling guide and principles for client.


src/components/Onboard/SerialPortOption.tsx
```diff
@@ -0,0 +1,91 @@
...
+          <div className=" border rounded-lg border-red-500 bg-red-50 pl-4 pt-5 pb-5 pr-4 w-6/12 ">
+            <div className="flex flex-row">
+              <XCircleIcon className="text-red-600 w-6 h-6" />
+              <h1 className="ml-4 text-base leading-6 font-normal text-red-600">
+                Serial port on {name}
+              </h1>
+            </div>
+            <h1 className="pl-6 pr-2 ml-4 text-sm leading-5 font-light text-red-600 mt-0.5">
+              Could not connect to serial port on {name}, failed with error
+              “Access Denied.”
```
uhuruhashimoto, line 69: Do we want this error to be hardcoded? The quotes imply that it's a lower-level error that propagated upwards. 

Are there reasons why the serial connection would fail that we can communicate to the user?

ajmcquilkin: Good catch, we can move the current `"Access Denied."` error into a prop that we pass in from the higher component before we can hook it up. This is related to #230

uhuruhashimoto (changes requested):
Approving with one blocker: please document your PR in the header comment, including the goal, implementation details and (if necessary) testing/output, as is done in other PRs.

uhuruhashimoto (approved):

PR #263, opened 2023-02-06
Samfydega/onboard functionality

samfydega:
In this PR, I set up the functionality for response coordinators to connect their nodes to the emergency client during onboarding. This ticket handles finding and connecting to the coordinator's node.

Goal: Implement functionality for the client onboarding page.
Implementation details: Followed styling guide and principles for client.


src/components/Onboard/SerialPortOption.tsx
```diff
@@ -1,37 +1,70 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
+import { useDispatch, useSelector } from "react-redux";
 import {
   RadioIcon,
   CheckCircleIcon,
   XCircleIcon,
 } from "@heroicons/react/24/outline";
+import { requestConnectToDevice } from "@app/features/device/deviceActions";
+import { selectDeviceConnected } from "@app/features/device/deviceSelectors";
```
ajmcquilkin, line 9: Nit: Import with `@features`

src/components/Onboard/SerialPortOption.tsx
```diff
@@ -1,37 +1,70 @@
...
-  connection_error,
+  isActive,
+  onSelection,
 }: ISerialPortOptions) => {
-  switch (state) {
+  const dispatch = useDispatch();
+  const isDeviceConnected = useSelector(selectDeviceConnected());
+  const statusOfConnect = useSelector(selectRequestStateByName(name));
+  const [portState, setPortState] = useState(state);
+  const [connectionError, setConnectionError] = useState("NULL.");
```
uhuruhashimoto, line 29: Is this descriptive enough as an initial state? I'm picturing the user getting a "NULL." error message and being confused. Not blocking PR on this though.

ajmcquilkin, line 29: What is the purpose of using string `"NULL"`? You could do either `""` or `null`

src/components/Onboard/SerialPortOption.tsx
```diff
@@ -1,37 +1,70 @@
...
+    } else {
+      if (statusOfConnect?.status === "FAILED") {
+        setPortState("FAILED");
+        setConnectionError(statusOfConnect.message);
+      }
+    }
+    if (isActive) {
+      setPortState("SUCCESS");
+    }
+  });
```
ajmcquilkin, line 48: When do you want this `useEffect` to run? This will run reach frame and likely cause a rerender cycle https://reactjs.org/docs/hooks-effect.html

src/components/Onboard/SerialPortOption.tsx
```diff
@@ -1,37 +1,70 @@
...
   name,
   state,
-  connection_error,
+  isActive,
+  onSelection,
 }: ISerialPortOptions) => {
-  switch (state) {
+  const dispatch = useDispatch();
+  const isDeviceConnected = useSelector(selectDeviceConnected());
+  const statusOfConnect = useSelector(selectRequestStateByName(name));
```
ajmcquilkin, line 27: What is this `name` field being used for? I was imagining you would use `useSelector(selectRequestStateByName(requestConnectToDevice.type))`

src/components/pages/OnboardPage.tsx
```diff
@@ -1,38 +1,34 @@
...
+import { useDispatch, useSelector } from "react-redux";
 import Hero_Image from "@app/assets/onboard_hero_image.jpg";
 import Meshtastic_Logo from "@app/assets/Mesh_Logo_Black.png";
 import SerialPortOption from "../Onboard/SerialPortOption";
 
-type PortState = "IDLE" | "LOADING" | "SUCCESS" | "FAILURE";
+import { requestAvailablePorts } from "@features/device/deviceActions";
+import { selectAvailablePorts } from "@app/features/device/deviceSelectors";
+
+type PortState = "IDLE" | "PENDING" | "SUCCESS" | "FAILED";
```
ajmcquilkin, line 10: We're now using the connection state from the `selectRequestStateByName` call, you might not need this.

src/components/pages/OnboardPage.tsx
```diff
@@ -1,38 +1,34 @@
...
-      state: "SUCCESS",
-    },
-    {
-      name: "port4",
-      state: "LOADING",
-    },
-  ];
+  const availableSerialPorts = useSelector(selectAvailablePorts());
+  const dispatch = useDispatch();
+  const [activeNode, setActiveNode] = useState(-1);
```
ajmcquilkin, line 20: Might just use `null` for this, makes life a bit easier in TS

ajmcquilkin (changes requested):
Primarily blocking on `useEffect` comment

src/components/pages/OnboardPage.tsx
uhuruhashimoto: Beginner question -- what's standard for hooks? empty string or `null`?

src/components/pages/OnboardPage.tsx
```diff
@@ -64,17 +60,28 @@ const OnboardPage = ({ unmountSelf }: IOnboardPageProps) => {
...
+                    }, 1500);
+                  }}
                 />
-              );
-            })}
+              ))
+            ) : (
+              <p className="text-base leading-6 font-normal text-gray-500 pl-40 pr-40 text-center">
+                {" "}
+                No ports detected.{" "}
```
sherry1iu, line 82: This isn't actually showing up when no nodes are plugged in.
image

sherry1iu (changes requested):

src/components/pages/OnboardPage.tsx
```diff
@@ -64,17 +60,28 @@ const OnboardPage = ({ unmountSelf }: IOnboardPageProps) => {
           </h1>
         </div>
         <div className="mt-10 flex flex-col">
-          <div className="flex flex-col gap-4" onClick={moveToMain}>
-            {mockPorts.map((port) => {
-              return (
+          <div className="flex flex-col gap-4">
+            {activePorts ? (
+              activePorts.map((port, index) => (
```
sherry1iu, line 65: There needs to be a way to refresh the page, maybe something taking the place of the "continue without radio" button" in the UI.

You can use something like in this:

https://upmostly.com/tutorials/how-to-refresh-a-page-or-component-in-react

src/components/pages/OnboardPage.tsx
```diff
@@ -64,17 +60,28 @@ const OnboardPage = ({ unmountSelf }: IOnboardPageProps) => {
           </h1>
         </div>
         <div className="mt-10 flex flex-col">
-          <div className="flex flex-col gap-4" onClick={moveToMain}>
-            {mockPorts.map((port) => {
-              return (
+          <div className="flex flex-col gap-4">
```
sherry1iu, line 63: add `overflow-y-auto`; the preview was overflowing for me

sherry1iu:
Looks good, just a few changes

src/components/Onboard/SerialPortOption.tsx
uhuruhashimoto: resolving; addressed below.

ajmcquilkin:
@sherry1iu @uhuruhashimoto @samfydega Added functionality changes, re-requesting review

src/components/Onboard/SerialPortOption.tsx
samfydega: fixed

src/components/Onboard/SerialPortOption.tsx
samfydega: This was to put a default connection error message. I've changed to "".

src/components/Onboard/SerialPortOption.tsx
samfydega: This was passed in as a prop to identify which component is currently being handled. I've updated it from name to your implementation.

src/components/pages/OnboardPage.tsx
samfydega: I believe this is happening because the map isn't null but there are no values. Should now be fixed.

src/components/pages/OnboardPage.tsx
samfydega: Added.

src/components/pages/OnboardPage.tsx
samfydega: Added.

src/components/pages/OnboardPage.tsx
samfydega: Updated.

src/components/pages/OnboardPage.tsx
samfydega: Fixed.

src/components/pages/OnboardPage.tsx
```diff
@@ -63,30 +108,67 @@ const OnboardPage = ({ unmountSelf }: IOnboardPageProps) => {
...
           </div>
         </div>
+
+        <button
+          type="button"
+          className="flex flex-row justify-center align-middle gap-4 mt-5"
+          onClick={() => requestPorts()}
+        >
+          <ArrowPathIcon className="text-gray-400 w-6 h-6 hover:cursor-pointer" />
+          <p className="my-auto text-gray-500">Refresh ports</p>
```
sherry1iu, line 141: Nitpick: this does not look like a button.

src/components/Onboard/SerialPortOption.tsx
```diff
@@ -70,8 +69,10 @@ const SerialPortOption = ({
               </h1>
             </div>
             <h1 className="pl-6 pr-2 ml-4 text-sm leading-5 font-light text-red-600 mt-0.5">
-              Could not connect to serial port on {name}, failed with error “
-              {connection_error}”
+              Could not connect to serial port on {name}, failed with error
+              &quot;
```
sherry1iu, line 73: When I tried this, it never actually showed; the loading would just go on indefinitely until the application was F5 refreshed or closed. I'd suggest adding a timeout

sherry1iu (approved):
Looks good! Two issues with this detailled above, but feel free to resolve them - I think it might be worth opening those as separate tickets.

ajmcquilkin (approved):
Worked with Sam, we should be good to go

PR #286, opened 2023-02-16
bug fix: sidebar size change

sherry1iu:
Sidebar shrank when on tabs that had w:full. Comparison below:

Before:
sidebar shrink prob

After:
sidebar shrink fix


ajmcquilkin (approved):
Good find!

PR #293, opened 2023-02-18
Implement dialogue popup w/o type blocking

### LINKED ISSUE #199: Rust production build failing
The Rust development build for the application (`pnpm run rust:dev`) is failing. All build failures appear to be type issues within `Settings.tsx` and `main.tsx`.

image

sherry1iu:
Implement dialogue popup using `radix-ui` library. 
Replaces `react-alert` functionality, and removes all references to it.

popup

Note: `Mapbox-gl` installed due to error in some other component

Closes #199 blocking on build due to types


package.json
```diff
@@ -17,13 +17,14 @@
...
+    "@radix-ui/react-dialog": "^1.0.2",
     "@reduxjs/toolkit": "^1.8.6",
     "@tauri-apps/api": "^1.1.0",
     "heroicons": "^2.0.13",
+    "mapbox-gl": "^2.12.1",
     "maplibre-gl": "^2.4.0",
     "prop-types": "^15.8.1",
     "react": "^18.2.0",
-    "react-alert": "^7.0.3",
     "react-alert-template-mui": "^1.0.7",
```
ajmcquilkin, line 28: Do we still need this package?

src/components/Map/MapIconButton.tsx
```diff
@@ -8,6 +9,40 @@ export interface IMapIconButtonProps {
...
+    <Dialog.Root>
+      {/* Icon on map */}
+      <Dialog.Trigger asChild>
+        <button className={`${className} default-overlay`}>{children}</button>
+      </Dialog.Trigger>
+
+      {/* Appears when clicked */}
+      <Dialog.Portal>
+        <Dialog.Overlay />
+        <Dialog.Content className="fixed top-1/4 left-1/3 rounded-lg bg-white drop-shadow-xl p-8">
```
ajmcquilkin, line 28: Why isn't this centered? Can we use `top-1/2 left-1/2 -translate-x-1/2 -translate-y-1/2`

sherry1iu: I did play around with these for a little, and thought that this looked better; would love to hear other opinions though.
`top-1/4 left-1/3`:
image

`top-1/2 left-1/2 -translate-x-1/2 -translate-y-1/2`:
image

ajmcquilkin: The difference in positioning becomes more challenging as the screen size gets larger, I would pretty strongly push towards centering it. Also might be worth putting a global shadow when the modal is active.

sherry1iu: Centered. I don't think adding the global shadow is necessary, since clicking anywhere outside the dialogue also closes it.

ajmcquilkin (approved):

PR #296, opened 2023-02-19
Reset app on device disconnect

### LINKED ISSUE #290: Notify user when serial device disconnects
Implement functionality to notify the user when their device disconnects, and show them UI to reconnect their device.

ajmcquilkin:
Currently, the application doesn't have UI for reconnecting to a disconnected device. This PR implements a flow that will automatically run the `disconnect_from_serial_port` command and then reload the UI's webview when a serial device loses connection with the application (a `BrokenPipe` error). This will allow the user to reconnect their device as if they just opened the app. This UX will be temporary until a reconnection flow can be established.

Closes #290

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -84,7 +84,7 @@ async fn connect_to_serial_port(
     let mut connection = SerialConnection::new();
     let new_device = mesh::device::MeshDevice::new();
 
-    connection.connect(port_name, 115_200)?;
+    connection.connect(app_handle.clone(), port_name, 115_200)?;
     connection.configure(new_device.config_id)?;
 
     let mut decoded_listener = connection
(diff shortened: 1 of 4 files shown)


sherry1iu:
If I'm understanding it right, this PR causes the app to reload upon losing connection, right? It's not happening for me:
broken_pipe



I have pulled the latest version of the branch:
image

ajmcquilkin:
Interesting, not sure why that's happening. I'll investigate on my Windows machine

ajmcquilkin:
@sherry1iu In theory this should be fixed, just pushed a Windows-specific error check. Linux throws a `BrokenPipe` error where Windows throws a `PermissionDenied` error, code is now checking for either.

sherry1iu (approved):
It works now, looks good!

One thing to note though - if someone has a bad usb connection, it could get really annoying having to go through the onboarding page each time. Not something to change with this PR, just something to note.

PR #298, opened 2023-02-19
Added multi-channel sending support

### LINKED ISSUE #270: Cannot send from channel other than primary
Currently, the Rust backend infrastructure doesn't allow us to send from a non-primary (index 0) channel, even when a non-zero channel is specified via the `send_text` command.

ajmcquilkin:
This PR removes the restriction in `requestSendMessage` that forces sending of messages on only the `PRIMARY` channel (index 0).

Closes #270


sherry1iu (approved):
Looks good!

PR #300, opened 2023-02-20
Bindings update

sherry1iu:
Ran `cargo test` to update bindings


src-tauri/bindings/ChannelMessagePayload.ts
```diff
@@ -2,4 +2,4 @@
 import type { TextPacket } from "./TextPacket";
 import type { WaypointPacket } from "./WaypointPacket";
 
-export type ChannelMessagePayload = { text: TextPacket };
+export type ChannelMessagePayload = { text: TextPacket } | { waypoint: WaypointPacket };
```
ajmcquilkin, line 5: This change will break builds, the UI expects a `.text` field on all messages for the time being

src-tauri/bindings/MeshChannel.ts
```diff
@@ -1,5 +1,5 @@
 // This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
-import type { Channel } from "./protobufs/Channel";
+import type { Channel } from "./protobufs\\Channel";
```
ajmcquilkin, line 2: The `\\` token will likely break UNIX paths

ajmcquilkin (changes requested):
Blocking on running this on macOS and Linux, we're likely going to remove `tsync` based on its inability to generate `enum` types

ajmcquilkin:
Following up on block, see #301

Issue #47, opened 2022-10-25
Set up GitHub Actions


ajmcquilkin:
Set up [GitHub Actions](https://github.com/features/actions) flows for the following cases:

- UI tests (`pnpm run ui:test`, unimplemented)
- Rust tests (`pnpm run rust:test`)

PR #306, opened 2023-02-22
Update eslint config

ajmcquilkin:
This PR updates the project's eslint configuration, and adds in a list of recommended VSCode extensions. Both the `ui:lint` and `ui:format` commands are now functional, and should be run on each PR manually before GitHub actions can be configured (see #47).


uhuruhashimoto (approved):
Info on PR that I was confused about: 
- Recommended extensions are in .vscode/settings.json. 
- This will not format automatically (just warn you/prevent merging if your code doesn't meet the standards).

Approving after discussion and clarification of these points :)

Issue #207, opened 2023-01-19
Fix issues with overlapping Rust -> TS proto types


ajmcquilkin:
Currently we use [RS-TS](https://github.com/Aleph-Alpha/ts-rs) to convert the generated Rust protobuf types into TypeScript types for consumption in the client (`/src-tauri/bindings`). This has the following issues:

1. This generation has to be run during test, which means this generation will happen during unrelated test runs.
2. The library we're using to generate these types doesn't handle overlapping types, meaning we have multiple `Role` objects that overwrite each other when trying to write into the `Role.ts` file.

This will likely require generating TypeScript types directly from the proto files with a library such as [protobuf-es](https://github.com/bufbuild/protobuf-es).

PR #309, opened 2023-02-22
Removed TS generation test suite

ajmcquilkin:
This PR removes the existing `TS_RS` type generation suite, per #207.


uhuruhashimoto (approved):
Approving. Discussion: we don't use our tests anyway, and removing types generation from them stops breaking things so we can use them again.

PR #326, opened 2023-02-27
Ajmcquilkin/integrate neighborinfo protos

uhuruhashimoto:
## Goal
This PR sets up the basics for our backend to receive `NeighborInfo` packets from firmware.

## Implementation
- This PR adds a handler for packets from the `NeighborInfo` (port 71).
- This PR adds a `NeighborInfoPacket` struct to pass the packet to the handler.
- This PR adds a `neighbors` field to the `MeshDevice` struct, intended to directly store `NeighborInfo` packets as they come in.
- This PR re-points the `protobufs` submodule to the `development` branch of the fork (https://github.com/uhuruhashimoto/protobufs), which contains the `NeighborInfo` protobuf, changes to `DeviceOnly` proto, and changes to `MeshPacket` (to pass in the info that will later be appended to the header). 

## Testing
Tested in hardware with two nodes with flashed firmware from `development` fork of https://github.com/uhuruhashimoto/firmware

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "src-tauri/protobufs"]
 	path = src-tauri/protobufs
-	url = https://github.com/meshtastic/protobufs
+	url = https://github.com/uhuruhashimoto/protobufs.git
+	branch = sim-working
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "src-tauri/protobufs"]
 	path = src-tauri/protobufs
 	url = https://github.com/uhuruhashimoto/protobufs.git
-	branch = sim-working
+	branch = development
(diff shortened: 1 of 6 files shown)


ajmcquilkin:
Looking over the code rn, but something to consider going forward is that any functionality we push should work even without modified firmware. Checking that's the case now.

ajmcquilkin (approved):
Looks good! Just a note -- there was an extra portNum println! macro call that we probably don't want to go into prod

uhuruhashimoto:
> Looks good! Just a note -- there was an extra portNum println! macro call that we probably don't want to go into prod

oh thanks! I blanked on removing that. 
We may be blocked on merging unless I approve it, since you pushed the last commit.

ajmcquilkin:
Do you want to merge it? I'll use admin if so

uhuruhashimoto:
> Do you want to merge it? I'll use admin if so

Yes, it should be ready to merge.

PR #329, opened 2023-02-28
Graph Topology Info to Backend

uhuruhashimoto:
## Goal
Create a persistent graph in the backend that we update based on the `MeshDevice` struct. This graph is updated constantly, and we'll clone it to pass it into our `state` module via `add_graph`.

## Implementation
- Added `MeshGraph` struct with a single `Graph` member
   - (can be easily cut back to a graph; I wasn't sure if we'd want to store an update timestamp or some other data along with the graph eventually).
- Added a mutex to lock the `MeshGraph` struct so we can also access it from commands
- Added a slightly cursed method of updating it within the `MeshDevice` struct 
   - we're passing the `MeshGraph` struct through the function chain (`main`->`handle_packet_from_radio` -> `handle_mesh_packet`), and updating it only if we found a packet for the `NeighborInfo` port.
- If we receive such a packet, constructing a new graph from our stored HashMap of `NeighborInfo`
- A rewrite of `init_edge_map` to check for dropped edges and take the most recently transmitted packet as consensus.

## Testing
Tested in hardware with two nodes. Correctly updating/recreating the graph struct at NeighborInfo packet reception. Note that SNR values are impossible (8.0 - 10.0), which is giving us NaN for weights. This should be investigated (the raw data and our conversion process). 

Also added a unit test for dropped edges.


uhuruhashimoto:
This PR is very naive. Here are a list of problems with it. 
1. It isn't modular. 
- It attaches a new struct to the `MeshDevice` struct, and in doing so muddies a lot of the `MeshDevice` logic. 
- This can be improved by a refactor, but it's easier to fix now. If anyone has a better implementation, I'm all ears.
2. It's expensive. 
- We recreate the graph from scratch (using an expensive edge creation function) every time we receive a packet, assuming that our entire `MeshDevice` store has changed. 
- This can be improved by dropping the sending frequency in the firmware and/or setting a time limit to trigger graph updates. 
3. It's prone to noise. 
- We currently aren't noise/sanity checking the data we create the Graph out of, which will cause us problems in the long run. 
- This can be improved by adding on another PR to check/limit SNR values.

On the bright side: 
It works. We now have a Graph to mess with.

src-tauri/src/data_conversion/distance_conversion.rs
```diff
@@ -12,10 +12,16 @@ use crate::mesh::device::MeshNode;
...
     let node_1_data = node_1.data;
     let node_2_data = node_2.data;
-    let node_1_pos = node_1_data.position.unwrap();
-    let node_2_pos = node_2_data.position.unwrap();
+    let has_node_1_pos = node_1_data.position;
+    let has_node_2_pos = node_2_data.position;
+    if has_node_1_pos.is_none() || has_node_2_pos.is_none() {
+        return default as f64;
+    }
+    let node_1_pos = has_node_1_pos.unwrap();
```
ajmcquilkin, line 23: We shouldn't unwrap this, if there's a chance of failure we should either return an error or an option

src-tauri/src/data_conversion/distance_conversion.rs
```diff
@@ -12,10 +12,16 @@ use crate::mesh::device::MeshNode;
...
 */
 pub fn get_distance(node_1: MeshNode, node_2: MeshNode) -> f64 {
+    let default = 0;
     let node_1_data = node_1.data;
     let node_2_data = node_2.data;
-    let node_1_pos = node_1_data.position.unwrap();
-    let node_2_pos = node_2_data.position.unwrap();
+    let has_node_1_pos = node_1_data.position;
+    let has_node_2_pos = node_2_data.position;
+    if has_node_1_pos.is_none() || has_node_2_pos.is_none() {
```
ajmcquilkin, line 20: This can be rewritten with a couple of (messy but type-safe) `if-let` statements ([link](https://doc.rust-lang.org/rust-by-example/flow_control/if_let.html)).

src-tauri/src/mesh/device/mod.rs
```diff
@@ -139,3 +139,30 @@ impl MeshDevice {
...
+        Self {
+            graph: Graph::new(),
+        }
+    }
+}
+
+impl Clone for MeshGraph {
+    fn clone(&self) -> Self {
+        Self {
+            graph: self.graph.clone(),
```
ajmcquilkin, line 165: Why did you implement this manually? This should be able to be derived, see the `MeshDevice` struct declaration

src-tauri/src/mesh/device/state.rs
```diff
@@ -316,3 +319,11 @@ impl MeshDevice {
         }
     }
 }
+
+impl MeshGraph {
+    pub fn update_graph(&mut self, device: MeshDevice) {
```
ajmcquilkin, line 324: Is `update_graph` a descriptive enough name for this? Not sure I get what this is doing

src-tauri/src/mesh/device/handlers.rs
```diff
@@ -193,9 +195,13 @@ impl MeshDevice {
                         .map_err(|e| e.to_string())?;
 
                     self.add_neighborinfo(NeighborInfoPacket {
-                        packet: packet.clone(),
-                        data: data.clone(),
+                            packet: packet.clone(),
```
ajmcquilkin, line 198: Nit: spacing

src-tauri/src/constructors/init/init_edge_map.rs
```diff
@@ -1,37 +1,72 @@
...
+                    }
+                }
             }
         }
     }
     snr_hashmap
 }
 
+// pulls corresponding neighbor from a hashmap of NeighborInfo packets
+pub fn get_other_edge_side_packet(
```
ajmcquilkin, line 53: Same here, not sure I understand what this function does. Also see `if-let` comment below

src-tauri/src/constructors/init/init_edge_map.rs
```diff
@@ -1,37 +1,72 @@
...
+        let node_1 = neighbor_packet.node_id;
+        for neighbor in &neighbor_packet.neighbors {
+            let node_2 = neighbor.node_id;
+            snr_hashmap.insert(
+                as_key(node_1, node_2),
+                (neighbor.snr as f64, neighbor.rx_time as u64),
+            );
+            // If the opposite neighbor is found on a recent packet, we take the most recent timestamp as fact.
+            let opposite_neighbor = get_other_edge_side_packet(node_2, node_1, &neighbors);
+            if opposite_neighbor.is_some() {
```
ajmcquilkin, line 22: See `if-let` below

src-tauri/src/constructors/init/init_edge_map.rs
```diff
@@ -1,37 +1,72 @@
...
                     snr_hashmap.insert(
-                        as_key(node_id, neighbor_id),
-                        (neighbor_packet.snr, neighbor_packet.timestamp),
+                        as_key(node_1, node_2),
+                        (neighbor.snr as f64, neighbor.rx_time as u64),
+                    );
+                } else {
+                    snr_hashmap.insert(
+                        as_key(node_1, node_2),
+                        (opposite_neighbor.snr as f64, opposite_neighbor.rx_time as u64),
```
ajmcquilkin, line 32: Nit: Could this if/else be done with a ternary or conditionally-assigned variable? Not sure it's worth having a completely different if/else block if only one arg is changing

src-tauri/src/main.rs
```diff
@@ -83,6 +83,7 @@ async fn connect_to_serial_port(
 ) -> Result<(), CommandError> {
     let mut connection = SerialConnection::new();
     let new_device = mesh::device::MeshDevice::new();
+    let graph = mesh::device::MeshGraph::new();
```
ajmcquilkin, line 86: I might place this in managed state per our discussion, which would allow graph modification from within commands

src-tauri/src/data_conversion/distance_conversion.rs
ajmcquilkin: Also see here for a more simplified syntax https://stackoverflow.com/questions/64293804/how-to-write-multiple-condition-in-if-let-statement

src-tauri/src/data_conversion/distance_conversion.rs
uhuruhashimoto: Switched to match statements to use data from the nested struct (unwrapping directly from references in the hashmap).

Docs are [here](https://doc.rust-lang.org/stable/book/ch18-03-pattern-syntax.html)

src-tauri/src/data_conversion/distance_conversion.rs
uhuruhashimoto: Same as above comment - swapped it for another match to check the struct fields.

src-tauri/src/constructors/init/init_edge_map.rs
uhuruhashimoto: swapped it with a match as well :)

The purpose of the function is to see if the other side of the edge exists in our `NeighborInfo` packets. I added explanatory comments and renamed the function `check_other_node_agrees`, which may not be any less confusing. Let me know if those two functions make sense now

src-tauri/src/mesh/device/handlers.rs
uhuruhashimoto: Thanks! fixed

src-tauri/src/constructors/init/init_edge_map.rs
uhuruhashimoto: Replaced with ternary (as close as Rust gets)

src-tauri/src/constructors/init/init_edge_map.rs
uhuruhashimoto: replaced with `match`

src-tauri/src/data_conversion/distance_conversion.rs
```diff
@@ -11,19 +11,28 @@ use crate::mesh::device::MeshNode;
...
-        node_1_pos.longitude_i as f64 * LON_CONVERSION_FACTOR,
-        node_1_pos.altitude as f64 * ALT_CONVERSION_FACTOR,
-        node_2_pos.latitude_i as f64 * LAT_CONVERSION_FACTOR,
-        node_2_pos.longitude_i as f64 * LON_CONVERSION_FACTOR,
-        node_2_pos.altitude as f64 * ALT_CONVERSION_FACTOR,
-    )
+pub fn get_distance(node_1: Option<&MeshNode>, node_2: Option<&MeshNode>) -> f64 {
+    let default_distance = 0.0;
+    match(node_1, node_2) {
+        (Some(node_1), Some(node_2)) => {
```
ajmcquilkin, line 17: Nit: If it makes it easier, you could even do an `if let (node1, node2) = (Some(node1), Some(node2)) { ... return value; } return default ` to replace the `match` statements with `if` statements

src-tauri/src/mesh/device/mod.rs
uhuruhashimoto: Our `graph_ds::Graph` (our struct, not petgraph itself) didn't implement it (and threw a ton of errors if I tried to add it implicitly). I've implemented `Clone` manually for the `Graph` instead of the `MeshDevice` struct. 

We still don't have `Serialize`, or `Deserialize`, which we'll need to send the graph to the frontend. I took a quick stab at implementing them but there's a lot of nesting (petgraph elements need to implement them too), so I'm going to put that in a different PR.

barkincavdaroglu: Oh I tried to implement Serialize/Deserialize but it was hell.

src-tauri/src/data_conversion/distance_conversion.rs
uhuruhashimoto: for some reason, they stay as Options when I do that.

```
if let (node1, node2) = (Some(node1), Some(node2)) {
   let node1_pos = node1.data.position;
   ...
}
```
doesn't let me access the node1.data field because it's an Option.
Leaving it as a match for that reason.

src-tauri/src/mesh/device/state.rs
uhuruhashimoto: renamed to `regenerate_graph_from_device_info`, and swapped the unwrap of the `meshdevice` for an `if let` destructuring pattern

src-tauri/src/mesh/device/mod.rs
uhuruhashimoto: Yeah, you can start it with a `serde::Serializer` and use `serialize_field` on the elements of the `Graph` but then you have to implement `Serialize` for nodes and edges of the `petgraph` and it gets messy. Seems out of scope for this PR, but should definitely be a follow up if we're sending our graph to the frontend with the `MeshDevice` info.

src-tauri/src/main.rs
uhuruhashimoto: Added an arc/mutex for the graph controlled by the app, unlocking it in `connect_to_serial_port`. Passed down the graph by reference through the `MeshDevice` functions.

ajmcquilkin (approved):
Looks super solid! @uhuruhashimoto and I spoke and live demoed this, graph is serializable and edges look realistic

src-tauri/src/mesh/device/mod.rs
uhuruhashimoto: Update: serializing graph now

PR #333, opened 2023-03-02
Added UI message acknowledgements

### LINKED ISSUE #332: Add ACK receipts for outgoing messages
Add a web-style acknowledgement indicator, which will tell the user when any node on the network acknowledges the message the user sent, or if there was a routing error during transmission.

ajmcquilkin:
This PR adds acknowledgement text for messages, as shown below.

Screenshot_20230302_150106

Closes #332


uhuruhashimoto:
This is a general thought - we may want to start considering more carefully the use of the word "state." It's used separately for the serial connection (where we have `connection_state` in main) and for analytics (there's a `state` module). In this case, adding a state to an existing message seems self-explanatory, and your struct is descriptively named. But we're ending up with a lot of local variables named `state`, which will become confusing eventually.

uhuruhashimoto (approved):
Looks great! Thanks for the clippy fixes. 

If tested in hardware, I say it's good to go.

PR #337, opened 2023-03-03
Command to Run Algos

uhuruhashimoto:
## Goal
Add ability to run algos from the frontend

## Implementation
- Added a state variable with a mutex
- Added a new function for the initialization of the Graph and State (which also calls clone on the arc - [docs](https://doc.rust-lang.org/std/sync/struct.Arc.html))
- Added serialization to the AlgoResults structs and enum
- fixed graph unit tests 

## Testing
- Client builds successfully 
- command can be invoked from frontend with `__TAURI_INVOKE__("run_algorithms", {bitfield:0b11011}).then(console.log).catch(console.error)`
- No testing with real graph for coherent results yet

### DIFF
--- a/src-tauri/src/analytics/algos_config.rs
+++ b/src-tauri/src/analytics/algos_config.rs
@@ -13,7 +13,7 @@ struct Activation {
 }
 
 pub struct Params {
-    pub params: HashMap<String, Box<dyn Any>>,
+    pub params: HashMap<String, Box<dyn Any + Sync + Send>>,
 }
 
 impl Params {
@@ -23,7 +23,7 @@ impl Params {
         }
     }
 
-    pub fn add_param(&mut self, key: String, value: Box<dyn Any>) {
+    pub fn add_param(&mut self, key: String, value: Box<dyn Any + Sync + Send>) {
         self.params.insert(key, value);
     }
 
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -59,19 +59,13 @@ fn main() {
             inner: Arc::new(async_runtime::Mutex::new(None)),
         })
         .manage(ActiveMeshDevice {
-            inner: Arc::new(async_runtime::Mutex::new(Some(
-                mesh::device::MeshDevice::new(),
-            ))),
+            inner: Arc::new(async_runtime::Mutex::new(None)),
         })
         .manage(ActiveMeshGraph {
-            inner: Arc::new(async_runtime::Mutex::new(Some(
-                mesh::device::MeshGraph::new(),
-            ))),
+            inner: Arc::new(async_runtime::Mutex::new(None)),
         })
         .manage(ActiveMeshState {
-            inner: Arc::new(async_runtime::Mutex::new(Some(
-                analytics::state::State::new(HashMap::new(), false),
-            ))),
+            inner: Arc::new(async_runtime::Mutex::new(None)),
         })
         .invoke_handler(tauri::generate_handler![
             get_all_serial_ports,
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -384,8 +384,8 @@ async fn run_algorithms(
         .ok_or("State not initialized")
         .map_err(|e| e.to_string())?;
 
-    state.set_graph(&graph);
-    state.set_bitfield(&bitfield);
+    state.add_graph(&graph);
+    state.set_algos(&bitfield);
     state.run_algos();
     state.get_algo_results()
 }
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -375,7 +375,7 @@ async fn run_algorithms(
 ) -> Result<AlgoStore, String> {
     let mut guard = mesh_graph.inner.lock().await;
     let mut state_guard = algo_state.inner.lock().await;
-    let graph = guard
+    let graph_struct = guard
         .as_mut()
         .ok_or("Graph not initialized")
         .map_err(|e| e.to_string())?;
@@ -384,8 +384,8 @@ async fn run_algorithms(
         .ok_or("State not initialized")
         .map_err(|e| e.to_string())?;
 
-    state.add_graph(&graph);
-    state.set_algos(&bitfield);
+    state.add_graph(&graph_struct.graph);
+    state.set_algos(bitfield);
     state.run_algos();
-    state.get_algo_results()
+    Ok(state.get_algo_results().clone())
 }
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -8,7 +8,7 @@ use analytics::algo_store::AlgoStore;
 use app::protobufs;
 use mesh::serial_connection::{MeshConnection, SerialConnection};
 use serde::{Deserialize, Serialize};
-use std::sync::Arc;
+use std::{sync::Arc, collections::HashMap};
 use tauri::{async_runtime, Manager};
 
 struct ActiveSerialConnection {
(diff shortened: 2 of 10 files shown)


src-tauri/src/main.rs
```diff
@@ -63,6 +68,11 @@ fn main() {
                 mesh::device::MeshGraph::new(),
             ))),
         })
+        .manage(ActiveMeshState {
+            inner: Arc::new(async_runtime::Mutex::new(Some(
+                analytics::state::State::new(HashMap::new(), false),
```
uhuruhashimoto, line 73: We're currently passing in the bool `save=false` to our timeline, so we can't get repeated snapshots. I have a couple of questions: 
1. I'm not sure why we're initializing a new State struct here and on line 109 - can we move it somewhere else? 
2. I'm not sure if we should enable save by default (it feels a bit messy since we haven't given the user the ability to specify a save directory or any direct control over the process). Should we pass the save flag in from the frontend?

src-tauri/src/main.rs
```diff
@@ -90,10 +101,12 @@ async fn connect_to_serial_port(
     mesh_device: tauri::State<'_, ActiveMeshDevice>,
     serial_connection: tauri::State<'_, ActiveSerialConnection>,
     mesh_graph: tauri::State<'_, ActiveMeshGraph>,
+    algo_state: tauri::State<'_, ActiveMeshState>,
 ) -> Result<(), CommandError> {
     let mut connection = SerialConnection::new();
     let new_device = mesh::device::MeshDevice::new();
     let new_graph = mesh::device::MeshGraph::new();
+    let state = analytics::state::State::new(HashMap::new(), false);
```
uhuruhashimoto, line 109: See above comment on `save`

src-tauri/src/analytics/algos_config.rs
```diff
@@ -23,7 +23,7 @@ impl Params {
         }
     }
 
-    pub fn add_param(&mut self, key: String, value: Box<dyn Any>) {
+    pub fn add_param(&mut self, key: String, value: Box<dyn Any + Sync + Send>) {
```
ajmcquilkin, line 26: Why are these parameters `Any`? Is there not a way to use an `enum` so that we don't have to manually specify `Send + Sync`?

src-tauri/src/main.rs
```diff
@@ -107,6 +120,7 @@ async fn connect_to_serial_port(
     let handle = app_handle.app_handle().clone();
     let mesh_device_arc = mesh_device.inner.clone();
     let mesh_graph_arc = mesh_graph.inner.clone();
+    let algo_state_arc = algo_state.inner.clone();
```
ajmcquilkin, line 123: Something that just came to mind (cleanliness, feel free to bump to a new issue or ignore), it might be worth adding a new command that initializes the graph and the algos. The `connect_to_serial` command ideally wouldn't initialize anything that it doesn't need to function, which I would say includes the algos and graph

src-tauri/src/main.rs
```diff
@@ -339,3 +367,26 @@ async fn get_node_edges(
 
     Ok(edges)
 }
+
+#[tauri::command]
+async fn run_algorithms(
+    bitfield: u8,
+    mesh_graph: tauri::State<'_, ActiveMeshGraph>,
+    algo_state: tauri::State<'_, ActiveMeshState>,
+) -> Result<AlgoStore, String> {
```
ajmcquilkin, line 376: Commands should return a `CommandState` error type, which implements `From<String>` so you might not need the `map_err` calls

uhuruhashimoto: oh fair point. The `AlgoResults` also include `Error` and `Missing/None` options for each algo, should I unpack that here before I return to the frontend?

ajmcquilkin: What do you mean unpack them? All `CommandError` ([here](https://github.com/ajmcquilkin/meshtastic-emergency-response-client/blob/84c947b5cec2ff4acd91f6b59441630fbe14823d/src-tauri/src/main.rs#L27)) does is wrap the error message to support the JS Error spec, it doesn't have any special handling on its own

barkincavdaroglu:
The discussion on saving data functionality is redundant for the time being. Algorithms that need prior data will use dummy data, but timeline must be kept until the the end, but this doesn't require saving data as the timeline is kept in memory.

src-tauri/src/main.rs
barkincavdaroglu: What do you mean by "repeated snapshots"?

ajmcquilkin: Good call, we should init state to `None` if it hasn't been initialized in this case.

src-tauri/src/main.rs
uhuruhashimoto: my bad - that wasn't a directly relevant observation. It should have been its own comment. Agree with you on `CommandError`.

src-tauri/src/analytics/algos_config.rs
uhuruhashimoto: The values of the hashmap are of type `Box<dyn Any>`. `Box` provides a pointer for (dynamic) heap allocation. Because all threads share the heap, this memory is not inherently threadsafe. 

Therefore, to make it threadsafe (and allow us to add a mutex/ARC to a `State` struct), I added the additional specifiers `Send` and `Sync` to the function definition. This isn't an enum - it means that any input of type `Any` must also implement `Send` and `Sync`. 
https://stackoverflow.com/questions/25649423/sending-trait-objects-between-threads-in-rust

uhuruhashimoto: Now, as to why we're using heap memory in the first place, I have no idea. @barkincavdaroglu, any chance you remember why?

src-tauri/src/main.rs
uhuruhashimoto: @barkincavdaroglu whatever we're saving if we pass in `save=true`

barkincavdaroglu: When save is true we just write timeline (which should already be in memory) to external space. No need to save data for MVP / technigala

src-tauri/src/analytics/algos_config.rs
barkincavdaroglu: We have to hardcode / know beforehand all the possible param types to use enum, which I didn't at the time. This was also easier to implement.

barkincavdaroglu: Feel free to change it though

src-tauri/src/main.rs
uhuruhashimoto: Resolving - `save` hardcoded to false for MVP based on above discussion.

uhuruhashimoto:
Testing problem: client only works for 0b11, not anything higher (`__TAURI_INVOKE__("run_algorithms", {bitfield:11}).then(console.log).catch(console.error)`). 

Anything higher gives me this error: 
Screen Shot 2023-03-03 at 5 12 26 PM

src-tauri/src/main.rs
uhuruhashimoto: Removed error mapping

src-tauri/src/main.rs
uhuruhashimoto: Resolving conversation by defaulting to `save` to `false` and initializing data structures in main as `None`.

src-tauri/src/main.rs
uhuruhashimoto: Resolving conversation by defaulting to `save` to `false` and initializing data structures in main as `None`.

src-tauri/src/main.rs
uhuruhashimoto: Added a new function, `initialize_graph_state`, and moved arc/mutex data structure initialization there.

src-tauri/src/analytics/algos_config.rs
uhuruhashimoto: Changing this implementation feels out of scope for this PR, because it involves a narrow refactor of code that was designed to be general. Because the current implementation works and makes sense, I'm leaving it as is.

ajmcquilkin (approved):
Looks solid! Just ran on my machine, algos didn't return anything bc I'm not connected but didn't see any issues during execution

PR #341, opened 2023-03-06
Waypoint edit

### LINKED ISSUE #216: Add non-functional on-map waypoint UI

### LINKED ISSUE #252: Add on-map waypoint creation

### LINKED ISSUE #253: Add waypoint map pin for existing waypoints

### LINKED ISSUE #325: Add on-map waypoint editing

### LINKED ISSUE #342: Add on-map waypoint deletion

sherry1iu:
Create, edit, delete waypoint; select and deselect
waypoint_crud

\

Expiry: set to a year out by default. On deletion, it sets to time deleted
image

Closes #216 (Add non-functional on-map waypoint UI)
Closes #253 (Add waypoint map pin for existing waypoints)
Closes #252 (Add on-map waypoint creation)
Closes #325 (Add on-map waypoint editing)
Closes #342 (Add on-map waypoint deletion)


sherry1iu:
### Notes:
image

Currently, the flow for creating a new waypoint on the map is that the waypoint dispatches when it's created. Then on clicking and editing it, it'll send out a new update. Ideally, we want to create a marker in `Map.tsx` without sending, to prevent misclick/cancel from sending unnecessary packets out. 

This might take a bit longer to implement, and probably won't happen before Tuesday.

To make this marker would require we loop through all of the nodes to find the one with the matching latitude/longitude - since the (Lat, Long) pair is the only distinct feature in a waypoint other than the ID, and the ID is a random value from the backend.

This would also require changing the logic of how the waypointMenu is shown - currently, we need an active node to show the menu; but without a blank node sent on the initial click, there can't be an active node to select. My current thought is to make another redux state for if a waypoint is being initialized, and have the editWaypointMenu shown for that.

src/components/Map/MapInteractionPane.tsx
```diff
@@ -1,26 +1,42 @@
...
-// When functionality is implemented, change component to MapIconButton.
-
 const MapInteractionPane = () => {
+  const dispatch = useDispatch();
+
+  const newWaypoint = useSelector(selectIsNewWaypoint());
+
+  // Toggles newWaypoint state, which allows for creation of new waypoints on map
+  const handleClickMapPin = () => {
+    dispatch(deviceSliceActions.setNewWaypoint(!newWaypoint));
```
ajmcquilkin, line 20: Not sure if the naming is clear here, what does the `newWaypoint` state represent?

src/components/Map/MapView.tsx
```diff
@@ -32,6 +45,30 @@ export const MapView = () => {
...
+
+  const handleClick = (e: MapLayerMouseEvent) => {
+    // Can only create new waypoint if the state is toggled
+    if (newWaypoint) {
+      const createdWaypoint: Waypoint = {
+        id: 0,
+        latitudeI: Math.round(e.lngLat.lat * 1e7), // Location clicked
+        longitudeI: Math.round(e.lngLat.lng * 1e7),
+        expire: Math.round(moment().add(1, "years").valueOf() / 1000), // Expires one year from today
+        lockedTo: 0, // Locked to me
```
ajmcquilkin, line 60: 0 signifies an open waypoint, not a locked waypoint. In this case, we would set the `waypoint.lockedTo` field to our `nodeId`. This being said, I would imagine open waypoints would be a good default

https://buf.build/meshtastic/protobufs/docs/main:meshtastic#meshtastic.Waypoint

src/components/Map/MapView.tsx
```diff
@@ -32,6 +45,30 @@ export const MapView = () => {
...
+        name: "New Waypoint",
+        description: "",
+        icon: 128529, // (-_-) emoji; placeholder
```
ajmcquilkin, line 63: I would set this as 0 by default

src/components/Map/MapView.tsx
```diff
@@ -107,6 +148,19 @@ export const MapView = () => {
             />
           ))}
 
+        {/* Visualize all waypoints */}
+        {waypoints
+          .filter(
+            (n) =>
+              (!!n.latitudeI && !!n.longitudeI) || // Shows pins with valid latitudes
```
ajmcquilkin, line 155: What does this condition block? This is fairly confusing to me, and I'm not sure this is doing what you're intending. Also note that it's exceedingly unlikely to get a location of exactly `(0., 0.)` due to GPS noise, even when located at `(0., 0.)`

src/components/Waypoints/Waypoint.tsx
```diff
@@ -0,0 +1,48 @@
...
+interface IWaypoints {
+  currWaypoint: Waypoint;
+}
+
+const Waypoints = ({ currWaypoint }: IWaypoints) => {
+  const dispatch = useDispatch();
+  const activeWaypointID = useSelector(selectActiveWaypointID());
+  const isSelected = currWaypoint.id === activeWaypointID;
+
+  const expired = currWaypoint.expire < moment().valueOf() / 1000; // Boolean
```
ajmcquilkin, line 23: Nit: can also use `Date.now() / 1000` to avoid moment lib invocation

src/components/Waypoints/Waypoint.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  const handleClick = () => {
+    console.log();
```
ajmcquilkin, line 26: Unneeded log

src/components/Waypoints/Waypoint.tsx
```diff
@@ -0,0 +1,48 @@
...
+    if (!isSelected) {
```
ajmcquilkin, line 27: Nit: `dispatch(deviceSliceActions.setActiveWaypoint(isSelected ? currWaypoint.id : null));`

src/components/Waypoints/WaypointIcon.tsx
```diff
@@ -0,0 +1,22 @@
...
+export interface IWaypointProps {
+  isSelected: boolean;
+  expired: boolean;
+}
+
+const WaypointIcon = ({ isSelected, expired }: IWaypointProps) => {
+  return (
+    <img
+      className={"drop-shadow-lg"}
+      src={expired ? "" : isSelected ? SelectedWaypoint : DefaultWaypoint}
```
ajmcquilkin, line 18: Why do we render the `WaypointIcon` at all if a waypoint is expired?

src/components/Waypoints/WaypointMenu.tsx
```diff
@@ -0,0 +1,126 @@
...
+    : null;
+
+  // Waypoint Utils, used when the buttons at bottom are clicked
+  const deleteWaypoint = useDeleteWaypoint();
+  const toggleEditWaypoint = useToggleEditWaypoint();
+
+  // Only show if there is an active waypoint
+  if (!activeWaypoint) {
+    return <></>;
+  } else {
```
ajmcquilkin, line 41: Nit: don't need an explicit `else` block since the `if` statement returns, meaning the only time we reach any code below the `if` statement is when `activeWaypoint` is not `null`

src/components/Waypoints/WaypointMenu.tsx
```diff
@@ -0,0 +1,126 @@
...
+    return (
+      <div className="absolute top-24 right-9 bg-white pt-5 pr-5 pb-3 pl-4 rounded-lg drop-shadow-lg w-80">
+        <div className="flex justify-between">
+          {/* Title */}
+          <h1 className="text-gray-600 text-2xl leading-5 font-semibold ">
+            {/* If waypoint has no title / empty string for a title*/}
+            {!waypointTitle || waypointTitle?.length == 0
```
ajmcquilkin, line 48: Nit: `waypointTitle?.length ? waypointTitle : "No Title"`

src/components/Waypoints/WaypointMenu.tsx
```diff
@@ -0,0 +1,126 @@
...
+          <h4 className="leading-6 font-semibold pt-2">Description</h4>
+          <h3 className="leading-5 font-normal py-1">
+            {!waypointDescription || waypointDescription?.length == 0
+              ? "No Description"
+              : waypointDescription}
+          </h3>
+        </div>
+
+        {/* Details */}
+        <h4 className="text-gray-500 text-base leading-6 font-semibold pt-2 pb-1">
```
ajmcquilkin, line 73: Headings should be used in order, here we're setting `h4` to be higher priority than `h3`.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#usage_notes

src/components/Waypoints/WaypointMenu.tsx
```diff
@@ -0,0 +1,126 @@
...
+                {waypointLat == null || waypointLong == null
+                  ? "No location data"
+                  : "(" + waypointLat + ", " + waypointLong + ")"}
+              </h3>
+            </div>
+            <button
+              type="button"
+              onClick={() =>
+                void writeValueToClipboard(
+                  waypointLat == null || waypointLong == null
```
ajmcquilkin, line 91: Nit: `waypointLat && waypointLong ? "(" + waypointLat + ", " + waypointLong + ")" : "No location data"`

src/components/Waypoints/WaypointMenuEdit.tsx
```diff
@@ -0,0 +1,174 @@
...
+    : null;
+  const waypointLong = activeWaypoint?.longitudeI
+    ? (activeWaypoint.longitudeI / 1e7).toString()
+    : null;
+
+  // Utils
+  const toggleEditWaypoint = useToggleEditWaypoint();
+  const handleSubmit: FormEventHandler = (e) => {
+    e.preventDefault();
+    console.log("Clicked send");
```
ajmcquilkin, line 38: Unnecessary log

src/components/Waypoints/WaypointMenuEdit.tsx
```diff
@@ -0,0 +1,174 @@
...
+        ...activeWaypoint,
+        name: waypointTitle ? waypointTitle : "",
+        description: waypointDescription ? waypointDescription : "",
+      };
+      dispatch(
+        requestNewWaypoint({ waypoint: updatedWaypoint, channel: channelNum })
+      );
+      dispatch(deviceSliceActions.setWaypointEdit(false));
+    } else {
+      console.log("Error: No active waypoint");
```
ajmcquilkin, line 55: Nit: `console.error`

src/components/Waypoints/WaypointMenuEdit.tsx
```diff
@@ -0,0 +1,174 @@
...
+    }
+  };
+
+  // Only display if there is a selected waypoint
+  if (!activeWaypoint) {
+    return <></>;
```
ajmcquilkin, line 61: See comment above, I would also recommend returning `null` instead of a fragment

src/components/Waypoints/WaypointMenuEdit.tsx
```diff
@@ -0,0 +1,174 @@
...
+            />
+          </h3>
+
+          <h4 className="text-gray-500 text-base leading-6 font-semibold pt-2 pb-1">
+            Location
+          </h4>
+          <div className="flex flex-col">
+            <div className="flex justify-between pb-1">
+              <div className="flex justify-start">
+                <h3 className="text-gray-500 text-base leading-5 font-normal ">
```
ajmcquilkin, line 116: Same as header issue above

src/features/device/deviceSlice.ts
```diff
@@ -1,19 +1,26 @@
...
 import type { MeshNode } from "@bindings/MeshNode";
+import type { Waypoint } from "@bindings/protobufs/Waypoint";
 
 export interface IDeviceState {
   device: MeshDevice | null;
   activeNode: MeshNode["data"]["num"] | null;
   availableSerialPorts: string[] | null;
   activeSerialPort: string | null;
+  activeWaypoint: Waypoint["id"] | null;
+  waypointEdit: true | false; // Controls if the waypoint edit menu, or the normal menu shows up on map
```
ajmcquilkin, line 12: Should define this as a `boolean` type

src/utils/waypoint.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+// This file contains utils for: Delete waypoint and Toggle editWaypoint
+
+// Delete waypoint
+const useDeleteWaypoint = () => {
+  const dispatch = useDispatch();
+  const activeWaypoint = useSelector(selectActiveWaypoint());
+
+  return () => {
+    console.log("Deleted waypoint");
```
ajmcquilkin, line 22: Unneeded log

src/utils/waypoint.tsx
```diff
@@ -0,0 +1,53 @@
...
+// Toggles state of editWaypoint
+const useToggleEditWaypoint = () => {
+  const dispatch = useDispatch();
+  const isWaypointEdit = useSelector(selectIsWaypointEdit());
+  return () => {
+    dispatch(deviceSliceActions.setWaypointEdit(!isWaypointEdit));
+  };
+};
+
+// TODO: const useSendWaypoint = (waypoint: Waypoint) => {}
```
ajmcquilkin, line 51: What is this for?

src/utils/waypoint.tsx
```diff
@@ -0,0 +1,53 @@
...
+  selectActiveWaypoint,
+  selectIsWaypointEdit,
+} from "@features/device/deviceSelectors";
+import { requestNewWaypoint } from "@features/device/deviceActions";
+import { deviceSliceActions } from "@features/device/deviceSlice";
+
+// This file contains utils for: Delete waypoint and Toggle editWaypoint
+
+// Delete waypoint
+const useDeleteWaypoint = () => {
```
ajmcquilkin, line 17: Is this a hook? If not, don't call it `use...`

https://reactjs.org/docs/hooks-custom.html#extracting-a-custom-hook

src/components/Waypoints/WaypointMenuEdit.tsx
ajmcquilkin: Might also be worth doing `console.warn` since this isn't a fatal error, up to you

ajmcquilkin:
UX looks solid! Past demo we're going to want to prioritize #346 to limit unnecessary network traffic, many of my comments are code-style and clarity notes

src/components/Map/MapView.tsx
sherry1iu: Set to 128205, round pushpin emoji (sually used as default emoji for maps) - not sure how Apple/Android/Webclient handles 0, but it should be able to handle this.

src/components/Map/MapView.tsx
sherry1iu: Good catch - I didn't realize latitudeI and longitudeI couldn't be null

src/components/Waypoints/WaypointIcon.tsx
sherry1iu: Since we get the list of waypoints from each node echoing it back, the expired waypoints are still in the list. Since the waypoint.map in `Map.tsx` isn't refreshing frequently, this prevents it from being rendered

src/utils/waypoint.tsx
sherry1iu: I figured that if we were going to be sending waypoints from the `WaypointEditMenu,` `MapView` (for new waypoints), and eventually `WaypointConfig`, it might be worth having a common one that they share instead of having a new function every time

ajmcquilkin: Sounds good!

src/components/Map/MapView.tsx
ajmcquilkin: I'm not concerned about compatibility necessarily since it's easy for the client to ignore it, moreso the UX of having a default emoji when the user doesn't explicitly set one

src/components/Map/MapIconButton.tsx
```diff
@@ -57,7 +57,7 @@ const MapIconButton = ({
     <button
       type={type}
       onClick={onClick}
-      className={`${className} default-overlay`}
+      className={`${className} rounded-lg border border-gray-200 shadow-lg bg-white`}
```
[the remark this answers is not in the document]
ajmcquilkin: Is it possible to override the bg color while keeping the remaining styling? This is just a code consistency issue, so feel free to resolve and/or ticket if this wouldn't work or be quick

src/components/Waypoints/WaypointIcon.tsx
ajmcquilkin: Okay, might be worth putting a ticket for blocking the render of this component when the waypoint has been deleted but seems fine for now

ajmcquilkin (approved):

PR #350, opened 2023-03-07
change ap and mincut results to string

uhuruhashimoto:
## Goal
Change return type of ap and mincut to string; return only those.

## Notes
We are now ignoring all but ap and mincut results (can add in to struct later as strings).

## Testing
Tested in hardware; got results.


ajmcquilkin (approved):

barkincavdaroglu:
LG

PR #351, opened 2023-03-07
Return algo result IDs as u32

uhuruhashimoto:
## Goal
Change returned ID from String to u32 (so we can look up nodes)

## Testing
Tested in hardware.
Screen Shot 2023-03-07 at 1 07 56 PM

## Note
This call may fail if the graph isn't fully populated yet.

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -392,7 +392,7 @@ async fn run_algorithms(
     state.run_algos();
     let algo_result = state.get_algo_results();
     // convert AP from a vector of NodeIndexes to a vector of IDs (strings)
-    let ap_string_vec: Vec<String> = match &algo_result.aps {
+    let ap_vec: Vec<u32> = match &algo_result.aps {
         APResult::Success(aps) => aps
             .iter()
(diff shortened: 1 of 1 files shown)


src-tauri/src/main.rs
```diff
@@ -415,18 +415,22 @@ async fn run_algorithms(
...
 pub fn node_index_to_key(
     nodeindex: &petgraph::graph::NodeIndex,
     graph: &graph::graph_ds::Graph,
-) -> Option<String> {
+) -> Option<u32> {
     graph.node_idx_map.iter().find_map(|(key, &val)| {
         if val == *nodeindex {
-            Some(key.clone())
+            let num = key.parse::<u32>();
+            match num {
```
uhuruhashimoto, line 430: if there's a better way to do this, let me know

ajmcquilkin: Why wouldn't you directly return `num` if it's of type `Option<u32>`?

uhuruhashimoto: unwrapped w ok

ajmcquilkin (approved):

Issue #352, opened 2023-03-07
Broken submodule def


D4rk4:
```
dark@hope{~/repos/meshtastic-emergency-response-client}:cat .gitmodules 
[submodule "src-tauri/protobufs"]
	path = src-tauri/protobufs
	url = https://github.com/uhuruhashimoto/protobufs.git
	branch = neighborinfo_module
dark@hope{~/repos/meshtastic-emergency-response-client}:git submodule update --init
fatal: No url found for submodule path 'protobufs' in .gitmodules
```

ajmcquilkin:
Sorry, forgot to update this in the README! We're currently working with a custom [protobuf repo](https://github.com/uhuruhashimoto/protobufs), meaning you'll need to re-initialize the `protobufs` submodule. I would try running `git submodule sync` and potentially `git submodule update --init --remote` if the first command doesn't work on its own.

Additionally we've added network analysis functionality using a [custom firmware](https://github.com/uhuruhashimoto/firmware), which is why we're using the custom protobufs. Feel free to try it out if you're interested!

D4rk4:
@ajmcquilkin problem in this commit - https://github.com/ajmcquilkin/meshtastic-emergency-response-client/commit/082f4a18f7fa73c765a07b9cfd9f0d8002686678#diff-413872f7b4b9114419fdb06b539b6e9c12423e180e770336ab85d068dee7dca5

Why protobufs added in the root of the repo? It's UI only a related commit!

D4rk4:
Also build is broken:
```
error: failed to run custom build command for `app v0.1.0 (/home/dark/repos/meshtastic-emergency-response-client/src-tauri)`

Caused by:
  process didn't exit successfully: `/home/dark/repos/meshtastic-emergency-response-client/src-tauri/target/debug/build/app-0c970fddb544ea97/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-changed=protobufs/
  Running: "/home/dark/.cargo/registry/src/github.com-1ecc6299db9ec823/protoc-bin-vendored-linux-x86_64-3.0.0/bin/protoc" "--include_imports" "--include_source_info" "-o" "/tmp/prost-buildcknn8U/prost-descriptor-set" "-I" "protobufs/"
```

ajmcquilkin:
> 

Interesting, I hadn't seen the protos get added to the root of the project. I think #355 removes this, I'll make sure in a sec.

[...]

PR #355, opened 2023-03-19
General Code Style Refactoring

ajmcquilkin:
This PR is (a very large) refactoring PR, not intended to change functionality, but intended to do the following:

1. Make code naming and syntax (variables, methods) more intuitive
2. Refactor for improved error handling and/or reliability
3. Improve maintainability of existing functionality

The following is a list of major changes made within this PR. Note that, while all tests are passing, this PR is worth checking out locally due to the large numbers of files touched.

- Refactor algorithm implementations to add a new "runner" pattern, where algorithms are managed by a struct. This struct is configured, and is then run. This pattern allows for a two-step flow, where algorithm runs can be configured separately from when they are run.
- Remove as many `unwrap` statements as possible, primarily improving `Option` and `Result` handling but also using explanatory `expect` statements where refactoring was too challenging.
- Refactored directory structure within `src-tauri/src/analytics` to group files with similar functionality
- Removed all calls to `pub(crate)` to reduce global namespace clutter
- Renamed `global_mincut` functionality to `karger_stein` to not conflict with `stoer_wagner` implementation
- Replaced bitfield algorithm flags with a dedicated optional `AlgorithmConfigFlags` struct
   - Updated algorithms sagas to accept new flag-based contract
- Misc. naming changes to clarify functionality and/or intent


barkincavdaroglu:
I think you can remove Karger Stein since its not deterministic. It's faster and finds optimal solution >= 90% of the time if number of nodes >= 6 but our graphs are not that big anyway so the optimality is more important. Up to you, we can also leave it up to the user.

ajmcquilkin:
Since you've already written the algo it might just be worth letting it sit for the time being, I can imagine a future in which it could be useful as a relatively fast approximation in the even that the deterministic algorithm takes a long time. I don't think we reference it anywhere at this point, but it's not really hurting the codebase.

ajmcquilkin:
Merging to attempt a fix for #352, will revert if needed.

PR #357, opened 2023-03-23
Add CLI option to programatically connect to devices on startup

### LINKED ISSUE #299: Add ability to define meshtstic port via command line option
Nice to have the ability to define mehtastic device port via the command line option.
Why?
- Application used by not technically trained user => sysop may define port in shortcut/shell script.
- Application running automatically when OS booted up and ready to action (kiosk mode)

ajmcquilkin:
Per suggestion by @d4rk4, this PR adds in the ability to programmatically connect to Meshtastic serial devices instead of manually selecting them during application launch. The syntax to do so follows the pattern `EXECUTABLE --port PORT_NAME` or `EXECUTABLE -P PORT_NAME`. If a port is specified, the application will store its name in Tauri state.

The UI layer then queries if the application has a port marked for automatic connection (the `request_autoconnect_port` command). If this command returns a valid port name, the UI layer will then connect to this port as if the user selected it manually (see `SerialConnectionPage.tsx`).

If connection fails, the application will display the serial connection UI as usual.

> **Note:** Running the application with CLI arguments requires the application to be built using the `pnpm run rust:build` command, since CLI parsing is not reliable in development with PNPM.

Closes #299

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -40,42 +40,41 @@ fn main() {
                     app.app_handle().manage(initial_device_state);
 
                     if let Some(port_arg) = args.get("port") {
+                        if port_arg.occurrences == 0 {
+                            return Ok(());
+                        }
+
                         if let serde_json::Value::String(port_name) = port_arg.value.clone() {
                             println!("Connecting to port: {:?}", port_name);
 
                             let handle = app.app_handle();
 
+                            let analytics_state = handle.state::<state::AnalyticsState>();
                             let connection_state = handle.state::<state::ActiveSerialConnection>();
                             let device_state = handle.state::<state::ActiveMeshDevice>();
                             let graph_state = handle.state::<state::NetworkGraph>();
 
-                            let (connection, device) =
-                                ipc::helpers::initialize_serial_connection_handlers(
-                                    port_name,
-                                    app.app_handle(),
-                                    device_state.inner.clone(),
-                                    graph_state.inner.clone(),
-                                )?;
-
-                            let connection_arc = connection_state.inner.clone();
-                            let device_arc = device_state.inner.clone();
-
                             // Need this to unlock async locks in sync function
-                            tokio::runtime::Builder::new_multi_thread()
+                            tokio::runtime::Builder::new_current_thread()
                                 .enable_all()
                                 .build()
                                 .unwrap()
                                 .block_on(async move {
-                                    {
-                                        let mut connection_guard = connection_arc.lock().await;
-                                        *connection_guard = Some(connection);
-                                    }
+                                    let _result_1 = ipc::helpers::initialize_graph_state(
+                                        graph_state.clone(),
+                                        analytics_state,
+                                    )
+                                    .await;
 
-                                    // Only need to lock to set device in tauri state
-                                    {
-                                        let mut device_guard = device_arc.lock().await;
-                                        *device_guard = Some(device);
-                                    }
+                                    let _result_2 =
+                                        ipc::helpers::initialize_serial_connection_handlers(
+                                            port_name,
+                                            app.app_handle(),
+                                            device_state,
+                                            connection_state,
+                                            graph_state,
+                                        )
+                                        .await;
                                 });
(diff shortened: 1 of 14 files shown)


src-tauri/src/main.rs
```diff
@@ -7,24 +7,81 @@ mod mesh;
...
-        })
-        .manage(state::NetworkGraph {
-            inner: Arc::new(async_runtime::Mutex::new(None)),
-        })
-        .manage(state::AnalyticsState {
-            inner: Arc::new(async_runtime::Mutex::new(None)),
+        .setup(|app| {
+            match handle_cli_matches(app, &mut inital_autoconnect_state) {
+                Ok(_) => {}
+                Err(err) => panic!("Failed to parse CLI args:\n{}", err),
```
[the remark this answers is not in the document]
ajmcquilkin: I should have clarified in a comment, the only time this will fail is if there's a misconfigured config file. Otherwise any given argument will just return 0 occurences.

src/components/pages/SerialConnectPage.tsx
```diff
@@ -8,10 +8,14 @@ import Meshtastic_Logo from "@app/assets/Mesh_Logo_Black.png";
 import SerialPortOption from "@components/Onboard/SerialPortOption";
 
 import {
+  requestAutoConnectPort,
   requestAvailablePorts,
```
uhuruhashimoto, line 12: Just out of curiosity, are there any limitations of which ephemeral ports we can use? Could we ever use a virtual port (like the ACM port)? Do we have to worry about port logic at all, or does that all happen under the hood?

src-tauri/src/main.rs
```diff
@@ -7,24 +7,81 @@ mod mesh;
...
 
 use std::sync::Arc;
-use tauri::async_runtime;
+use tauri::{async_runtime, Manager};
+
+fn handle_cli_matches(
+    app: &mut tauri::App,
+    inital_autoconnect_state: &mut state::AutoConnectState,
+) -> Result<(), String> {
+    match app.get_cli_matches() {
```
uhuruhashimoto, line 16: What happens if a user doesn't specify a port and has no default saved? Also, why are we saving the default port if we don't persist state across app launches?

src/components/pages/SerialConnectPage.tsx
ajmcquilkin: In theory you could point this towards any file or directory, although the autoconnection would then fail, pushing them back to the default connection screen. Not sure how that failure would log in terminal though

ajmcquilkin: In theory you could point this towards any file or directory, although the autoconnection would then fail, pushing them back to the default connection screen. Not sure how that failure would log in terminal though

src-tauri/src/main.rs
ajmcquilkin: Fair point, the reason we're storing the autoconnect port in Tauri state is so that we can persist it across UI restarts. I also couldn't find an elegant way to immediately connect to a serial port in the `startup()` method (since it's not `async` :/). If they don't specify a port, the `port` argument will be passed with `{ occurrences: 0 }`, meaning the method will just ignore it and not set the state field.

uhuruhashimoto: Gotcha, makes sense. thanks for the explanation!

uhuruhashimoto (approved):

uhuruhashimoto (approved):
Just looked it over again and everything looks great.
- The backend async handling makes sense (including the refactoring you've done to put everything in one place).
- The only thing I was confused on was the error flow, which is a bit opaque at first glance. In terms of UX, it makes sense to me to perform a search and if no options are found, to default to normal setup flow. This choice isn't immediately apparent from the code, but it's well documented in the PR and issue conversation. Thoughts on noting it in the config file description?

Approved - feel free to merge.

ajmcquilkin:
Which config file would you be looking to note this in? Definitely on board with documenting it but I'm not sure I'm understanding what you think would be most effective

PR #358, opened 2023-04-03
Project Rebrand to "Meshtastic Network Management Client"

ajmcquilkin:
This PR is a tracking PR for a potential project rename.

As we've discussed internally, our project currently doesn't focus specifically on the emergency response scenario, and if we were to double down on this area of focus, we would run the risk of alienating potential users that would find our network and offline management features useful.

From this, I'm proposing we rename the project from "Meshtastic Emergency Response Client" to "Meshtastic Network Management Client." Within our internal conversations I've heard a lot of talk around making sure that Meshtastic networks are reliable and ensuring users can know enough about their network state to maintain working connections. This would certainly not preclude any work that would assist emergency response workers (much to the contrary I would argue), but we could then expand our reach to any potential application in which it would be advantageous to deploy a Meshtastic network.

If we were to go through with this change, this would entail the following changes:

1. Changing the name of the repository
2. Updating all references of "Emergency Response" to "Network Management" within the project's READMEs
3. Same update within all UI code

I'm not sure where I stand on the logo, since while the plus originally represented the US medical symbol, I could also see a framing where the plus represents the augmented capabilities of the client we're offering. Either way, I would argue that any logo change should be its own separate PR.

Tagging all members of the original project team as reviewers for input.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 <div align="center">
   <h1 align="center">Meshtastic Network Management Client</h1>
-  <p align="center">An unofficial Meshtastic desktop client, allowing simple, offline deployment and administration of a mesh communication network.</p>
+  <p align="center">An unofficial Meshtastic desktop client for analyzing and managing large-scale, low-bandwidth mesh networks</p>
 </div>
 
 <p align="center">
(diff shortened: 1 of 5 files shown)


barkincavdaroglu:
Do we still want Meshtastic's name? The change from "Emergency Response" to "Network Management" sounds sound, but can we not future proof and make it just a variant of "Radio Network Management Client"?

barkincavdaroglu (approved):

ajmcquilkin:
> Do we still want Meshtastic's name? The change from "Emergency Response" to "Network Management" sounds sound, but can we not future proof and make it just a variant of "Radio Network Management Client"?

That's a fair point, we could even call it "Mesh Network Management Client" if we completely wanted to move away from the Meshtastic brand. This being said a move away from Meshtastic would also mean we would have to ditch the logo, which would be a whole other design process.

uhuruhashimoto:
> > Do we still want Meshtastic's name? The change from "Emergency Response" to "Network Management" sounds sound, but can we not future proof and make it just a variant of "Radio Network Management Client"?
> 
> That's a fair point, we could even call it "Mesh Network Management Client" if we completely wanted to move away from the Meshtastic brand. This being said a move away from Meshtastic would also mean we would have to ditch the logo, which would be a whole other design process.

I like "Mesh Network Management Client" as a general summary, but it feels wordy as a replacement to Meshtastic. Also mixed on switching out entirely, unless we have a good alternative - I'll try to think of some!

sherr-berrt:
I'm also good with "Mesh Network Management Client!" I think we also discussed asking to integrate it with official Meshtastic down the road, so this might still be temporary. Also would we want something that points specifically at being a desktop client?

sherr-berrt:
> To be honest I hadn't really thought much about adding desktop branding in the name, is this something that other apps typically do? Would be open if that's a common naming pattern

Fwiw I was thinking of the other meshtastic apps, like the Meshtastic android and Meshtastic web client etc, no clue if it's common practice

uhuruhashimoto:
> > To be honest I hadn't really thought much about adding desktop branding in the name, is this something that other apps typically do? Would be open if that's a common naming pattern
> 
> Fwiw I was thinking of the other meshtastic apps, like the Meshtastic android and Meshtastic web client etc, no clue if it's common practice

I think Desktop is used occasionally; e.g. Azure Desktop, GitHub Desktop. I could vaguely see "Desktop Mesh Network Management"/"Mesh Network Management Desktop" -- which are four words I would never have applied to each other otherwise.

ajmcquilkin:
My takeaways from this: 

1. People are okay moving away from a specific focus on emergency response and more towards network management
2. Staying with "Meshtastic" branding for the time being would be simpler until we have a strong reason to switch branding, which would also be a straightforward switch from "Meshtastic" to "Mesh"
3. Still in discussion on how/whether to include "Desktop" in the name of the client, in which case we would want to follow existing naming patterns found in industry and within the Meshtastic community

Given these points I'm going to merge this PR to get the primary README changes into prod, but I'm personally very open to more discussion within an issue if that's something anyone feels passionate about.

PR #360, opened 2023-04-09
Fail serial connection flow on configuration timeout

ajmcquilkin:
This PR changes the condition for a successful serial connection within the `SerialConnectPage` flow. Currently all that is required to "successfully" connect to a serial device is for the `connect_to_serial_port` command to succeed. The issue with this is that, in the event that receiving configuration data from the serial device fails (device config, module config, channels), this failure will be silent to the user. This PR changes this "success" requirement to receiving a `ConfigCompleteId` packet from the serial device within a set timeout period (currently 1500ms). If this timeout is reached, the device will fail to connect and this failure will display on the `SerialConnectPage` UI in a user-readable manner.

Screenshot_20230409_144339

This change relies on the `SerialDeviceStatus` struct, shown below:

```rust
pub enum SerialDeviceStatus {
    Restarting,   // unused
    Disconnected, // no attempt or failure to connect
    Connecting,   // connection initialized, not yet configured
    Reconnecting, // unused
    Connected,    // successful serial connection and device configuration, UI notified
    Configuring,  // configuration in process
    Configured,   // configured but UI not yet notified
}
```
When a device is first created, it will start in the `Disconnected` state. When a serial connection is initialized to this device, it will move to the `Connecting` state. Once configuration is attempted, the device will move to the `Configuring` state and a timeout (currently 1500ms) will start. If the device does not receive a `ConfigCompleteId` packet within this timeout window, the device will notify the UI of the failure and will move back into the `Disconnected` state. If this packet is received within the timeout window, the device will move into the  `Configured` state. It will stay in this state until the UI layer is notified via the `"configuration_status"` event, at which point it will transition into the `Connected` state.

This PR is dependent on #359.

### DIFF
--- a/src-tauri/src/ipc/commands.rs
+++ b/src-tauri/src/ipc/commands.rs
@@ -120,7 +120,7 @@ pub async fn send_text(
         channel,
     )?;
 
-    events::dispatch_updated_device(app_handle, device.clone()).map_err(|e| e.to_string())?;
+    events::dispatch_updated_device(&app_handle, device.clone()).map_err(|e| e.to_string())?;
 
     Ok(())
 }
@@ -197,7 +197,7 @@ pub async fn send_waypoint(
         channel,
     )?;
 
-    events::dispatch_updated_device(app_handle, device.clone()).map_err(|e| e.to_string())?;
+    events::dispatch_updated_device(&app_handle, device.clone()).map_err(|e| e.to_string())?;
 
     Ok(())
 }
--- a/src-tauri/src/ipc/events.rs
+++ b/src-tauri/src/ipc/events.rs
@@ -2,8 +2,10 @@ use crate::mesh;
 use log::{debug, trace};
 use tauri::Manager;
 
+use super::ConfigurationStatus;
+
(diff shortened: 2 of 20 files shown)


uhuruhashimoto (commented):
Cool use of an enum to create a state machine! Are we sure that all states have defined error paths (e.g. that there's nowhere we could get stuck in connection flow?) If the answer is yes, this looks great to merge.

ajmcquilkin:
The default error state is always the disconnected state if that's what you mean! In the event any part of this fails, the UI will be instructed to disconnect the device

uhuruhashimoto:
> The default error state is always the disconnected state if that's what you mean! In the event any part of this fails, the UI will be instructed to disconnect the device

Gotcha. I phrased that poorly -- I don't think I had a fully formed question, I wanted to make sure I wasn't missing any bugs in the transition logic. 

So I just went through the logic again in detail and it looks great! Approving.

uhuruhashimoto (approved):
Looks great - the central states are very helpful. Approving!

PR #371, opened 2023-05-04
Update Types Generation

### LINKED ISSUE #207: Fix issues with overlapping Rust -> TS proto types

ajmcquilkin:
This PR uses an [updated version](https://github.com/ajmcquilkin/specta/tree/ajmcquilkin/ts-namespaces) of the Rust [Specta library](https://github.com/oscartbeaumont/specta) to generate TypeScript types at runtime from backend Rust types. This guarantees that the TypeScript types will not fall out of sync with the protobuf or Rust types when developing UI code.

This is done by removing the previously generated TypeScript bindings types and calling the `specta::export::ts(...)` function in the Rust application's `main` function. This will generate the TS types each time the application is started.

In the future  ideally we would do this at build time, but for the time being runtime is a good MVP.

Closes #207


uhuruhashimoto:
@ajmcquilkin nice job!

PR #434, opened 2023-08-13
Graph Infrastructure Overhaul

### LINKED ISSUE #287: Allow users to configure frequency of NeighborInfo packet

ajmcquilkin:
This PR overhauls the graph management infrastructure of the client. Generally these changes are intended to support the new neighbor information infrastructure being developed into the firmware, but this PR is also intended to generally clean up the legacy graph code.

This PR does the following:
- Updates the protos tag to `v2.2.0`
- Implements a `MeshGraph` struct that wraps the `Graph` struct
- Includes an edge and node timeout priority queue (inverted, low number -> high priority)
- Adds a handling flow for the new `NeighborInfo` packet, which adds nodes and edges to the graph and queues them to time out based on the received `node.node_broadcast_interval_secs` field
- Updates the timeout time of nodes and edges when new `NeighborInfo` packets are received, and due to the firmware waiting `2 * node.node_broadcast_interval_secs` to time out nodes, nodes and edges will only time out when they haven't been heard from by the device in a significant amount of time
- Adds a polling graph worker thread to purge timed out nodes and edges every 60s
- Removes unused or legacy code (`.unwrap` -> `Result` specifically)
- Added a new "Debug" tab group in the application sidebar
- Add new "Application Graph" debug tab visualizing the current neighbor information held by the client
- Added a "Neighbor Info" module configuration page
- Fixed a bug where a position would be considered invalid if either (not both) the lat or long coordinate was 0.0
- Listed neighbors for map active node pane

Closes #287


src-tauri/src/device/mesh_graph.rs
```diff
@@ -0,0 +1,402 @@
...
+
+        // Convert the NeighborInfo data into a GraphNode
+        let node_index = self.add_node(data.into());
+
+        for neighbor in neighbors {
+            // Implicitly convert the Neighbor data into a GraphNode
+            // let neighbor_index = self.add_node(neighbor.into());
+            let neighbor_index = self.add_node(GraphNode::new(
+                neighbor.node_id,
+                // TODO this should be used once the firmware doesn't zero the internal value
```
GUVWAF, line 109: I think the way to go is to loop through all neighbors that this node currently has, and remove those that are not in the newly incoming `NeighborInfoPacket`. That means they timed out in the firmware.

jp-bennett:
So far it looks like nodes only show up on the map when a Neighbors packet is received from that node. Simple location updates should be enough to put a node on the map.

PR #476, opened 2025-01-25
Fixed the message screen

RCGV1:
Acknowledge and transmit messages were previously flipped


turnrye (approved):
Thanks for fixing this bug!

PR #478, opened 2025-02-07
Fix some linter errors

matthewCmatt:
Looks like the repository currently doesn't pass the Biome linter.

This MR:
- Upgrades Biome to latest version (1.9.4)
  - See `package.json` and lock file
- Ignores the [`lint/suspicious/noReactSpecificProps`](https://biomejs.dev/linter/rules/no-react-specific-props/) rule, as we are using React!
  - See `biome.json`
- Runs [`biome lint --write`](https://biomejs.dev/reference/cli/#biome-lint) against the whole repo to fix safe issues
  - This is the bulk of the files touched, so please review knowing anything other than above was auto-generated
  - Looks like most of these are:
    - Renaming variables to fit Biome's standards
    - Specifying type imports
    - Specifying the module for some globals like `parseInt()`

Prior to this MR, the linter gives 134 errors and 1001 warnings.
After, it gives 5 errors and 697 warnings.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -33,7 +33,8 @@
         "noAccumulatingSpread": "off"
       },
       "suspicious": {
-        "noDoubleEquals": "warn"
+        "noDoubleEquals": "warn",
+        "noReactSpecificProps": "off"
       },
       "style": {
         "noImplicitBoolean": "off"
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "uuid": "^9.0.0"
   },
   "devDependencies": {
-    "@biomejs/biome": "^1.4.1",
+    "@biomejs/biome": "^1.9.4",
     "@tauri-apps/cli": "^1.1.1",
     "@types/jest": "^29.2.0",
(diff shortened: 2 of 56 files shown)


matthewCmatt:
Update: PR is created partially in error.

I was using too new of a PNPM version, which caused the lockfile to get ignored and dependencies (namely tsc and biome) were upgraded by PNPM.

From the [PNPM 10 release notes](https://github.com/pnpm/pnpm/releases/tag/v10.0.0):
> Removed conversion from lockfile v6 to v9. If you need v6-to-v9 conversion, use pnpm CLI v9.

The lockfile on `main` is v6:
> lockfileVersion: '6.0'

Downgrading to PNPM 9.15.2 (the last 9 release) allows the lockfile to be respected. With this, the build passes all linter rules and TSC.

We should note that PNPM 10+ is incompatible with the current lockfile/package.json somewhere prominent (README?), or pin our current package dependencies to stop PNPM from upgrading dependencies past what we support. I'll create an issue for this.

Issue #429, opened 2023-07-27
[Feature Request]: Updated macOS icon


rcarteraz:
This is a very trivial, low-priority request and I'm not even sure how feasible it is considering this is a multi-platform application. But, currently the icon for the macOS application is a circle, which stands out against the square with rounded edges that's predominantly used by macOS apps. If possible, I'd like to see the icon shape match the system. 

Here's a view of the icon with standard ones:
<img width="242" alt="Screenshot_2023-07-27_at_7 24 51_AM" src="https://github.com/ajmcquilkin/meshtastic-network-management-client/assets/6912600/b684487f-6b0c-47bb-9528-bea268589968">

Some resources: 
https://developer.apple.com/design/human-interface-guidelines/app-icons#macOS-app-icon-sizes
https://developer.apple.com/design/resources/

mc-hamster:
Let's give this a different icon all together, still with the //\ logo, but different enough not to confuse it with the iOS.

mc-hamster:
it needs different branding. someone should be able to look at it and know what it is right away.

PR #482, opened 2025-02-17
Update MacOS Icon

matthewCmatt:
Aims to solve #429

# Changes
- Changes the Tauri `productName` from "meshtastic-network-management-client" to "Meshtastic"
- Changes the MacOS icon to a new, rounded square icon.

Screenshot 2025-02-16 at 8 41 06 PM

Below is a short video summarizing these changes.

https://github.com/user-attachments/assets/90b095c5-2ff9-47fd-9174-96891b612d0d

The leftmost icon is the current NMC icon, mentioned in #429.
The center icon is the iOS Meshtastic app running on MacOS.
The rightmost icon is the new icon, with updated `productName` appearing on hover

# Details
### Is it differentiable from the iOS app?
I think the typical user won't install both apps, so this isn't a huge issue. If the user does have both, they can differentiate by the background color, or by opening the apps. Advanced users will figure it out however.

### Does this update Windows/Linux icons?
No. I wasn't sure what shape would be good for these, so I left them as-is for now. Maybe just the same updated design, in a circle?

### What is `productName`?
This is the name displayed to the user:
- In the task bar/dock
- In the Windows start menu/MacOS launch pad
- In the process/task manager

It is not:
- The title of the window, which remains the full app name:
Screenshot 2025-02-16 at 8 40 17 PM

### How did you create the icon?
1. Grab the Large White icon from the `design` repo https://github.com/meshtastic/design/blob/master/logo/svg/Mesh_Logo_White.svg
2. Drop it into an icon generator https://icon.kitchen/
3. Add a margin of 13% and a gradient using the background colors from the NMC app

Note: The rest of the icons in `src-tauri/icons` were generated using the `tauri icon` CLI. This one was not. On subsequent icon updates, this should be noted.


matthewCmatt:
@rcarteraz Thanks for taking a look at this so quickly.

Changed the productName to match the current branding ("Meshtastic Network Management Client").

I'd love to have a discussion about shortening this name in general (maybe just dropping "Client"?), but that's a matter for another PR.

caveman99 (approved):

caveman99:
Just something worth noting. I really like the white-on-black logo design introduced here and would also like to see that on the other platforms and in the client itself. The original logo design including the little plus sign to the upper left has since been re-used for meshtastic solutions. I think it fits there much better (meshtastic with added support, which the plus denotes).

matthewCmatt:
@rcarteraz Yeah, no problem.

Here's a quick draft of a Linux/Gnome-style logo. I tried to keep in line with [the GNOME App Icon guidelines](https://developer.gnome.org/hig/guidelines/app-icons.html):

gnome

Thoughts on this vs a circular version?

matthewCmatt:
@rcarteraz This one's ready for review.

Added a circular logo for Windows/Linux, and switched out the logos in the app to remove the +

PR #488, opened 2025-02-22
Upgrade Vite

matthewCmatt:
This MR upgrades to the latest version (v6)

This repo uses a fairly old version of Vite. It would be nice to switch to using Vitest for front-end testing, but it's only supported in >Vite 5.

 I looked through the Vite migration guides from [v4 -> v5](https://v5.vite.dev/guide/migration) and [v5 -> v6](https://vite.dev/guide/migration) and did a skim of the repo. Didn't find any incompatibilities on my search.


danditomaso (approved):
This looks good!

PR #491, opened 2025-03-02
ci: Add Devcontainer to project

rstuhlmuller:
Adding devcontainers makes keeping versions constant easier, and makes the development lifecycle much easier as well.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,15 @@
+FROM mcr.microsoft.com/devcontainers/typescript-node:1.1-22
+
+# Install Xvfb
+RUN apt-get update 
+RUN apt-get install -y \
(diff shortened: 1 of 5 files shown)


matthewCmatt:
Thanks for the PR! Really like reducing friction for contributions, so dev containers would be nice to have.

It looks like this PR also aims to add Dependabot. Would you mind backing that out and open another PR? Just to keep discussion/tracking atomic.

Are you able to run this and forward the Tauri window out to non-linux OS's following the README? I'm running `pnpm rust:dev` on MacOS and getting:
```
(meshtastic-network-management-client:12093): Gtk-WARNING **: 22:33:35.678: cannot open display: :99
```

If this is expected, could you add some more details for setup, or state the supported platforms for the dev container?

Issue #472, opened 2024-10-13
[Feature Request]: Docker support


Th3H0ff:
I managed to put together a Dockerfile that does compile but stops on some Tauri error. I haven't run across Tauti before and have a tough time proceeding further. I did some troubleshooting with the help of ChatGPT but got stuck on some Tauri configuration files .

The Dockerfile I've created

```
FROM node:18-slim

# Install system dependencies required by Tauri and Rust
RUN apt-get update && \
    apt-get install -y curl build-essential libssl-dev libwebkit2gtk-4.0-dev libgtk-3-dev \
    libayatana-appindicator3-dev librsvg2-dev && \
    rm -rf /var/lib/apt/lists/*

# Install Rust and Cargo
RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
ENV PATH="/root/.cargo/bin:${PATH}"

# Install the Tauri CLI using Cargo
RUN cargo install tauri-cli && \
    echo "Tauri installed, checking version..." && \
    ls /root/.cargo/bin && \
    /root/.cargo/bin/cargo-tauri --version

# Install PNPM globally
RUN npm install -g pnpm

# Set the working directory in the container
WORKDIR /usr/src/app

# Copy the package.json and pnpm-lock.yaml to the container
COPY package.json pnpm-lock.yaml ./

# Install dependencies using PNPM
RUN pnpm install

# Copy the rest of the application code to the container
COPY . .

# Build the frontend using Vite
RUN pnpm run ui:build

# Expose the port on which the application will run
EXPOSE 5173

# Set the environment variable for running in development mode
ENV NODE_ENV=development

# Command to run the Rust development server using PNPM
CMD ["cargo-tauri", "dev", "--", "--no-default-features"]
```

matthewCmatt:
Hey @Th3H0ff, could you explain a bit more about this suggestion? Are you thinking about dev containers?

If you're looking for a Meshtastic client to be hosted from a container, consider [the web client](https://github.com/meshtastic/web). If you want to run Meshtastic client commands from inside a docker container in a more automated environment, [the Python library/CLI](https://github.com/meshtastic/python) would probably be easier

I'm not sure if the above have already have container files made though. Would be a great contribution if not!

matthewCmatt:
Partially addressed by #493. Could use a standalone Dockerfile for non-VSCode development or CI.

Issue #471, opened 2024-10-12
[Bug]: pnpm run rust:dev fails with `Found argument '/usr/bin/node-20' which wasn't expected, or isn't valid in this context`


alistair23:
## Describe the bug

I have followed the steps from https://github.com/meshtastic/network-management-client?tab=readme-ov-file#installation for Fedora 40, but when running `pnpm run rust:dev` I get the following error

```shell
$ pnpm run rust:dev

> meshtastic-network-management-client@0.3.1 rust:dev /var/home/alistair/sources/network-management-client
> tauri dev -- --no-default-features --

error: Found argument '/usr/bin/node-20' which wasn't expected, or isn't valid in this context

USAGE:
    cargo-tauri [OPTIONS] <SUBCOMMAND>

For more information try --help
```

## To Reproduce

1. Try to build following steps from https://github.com/meshtastic/network-management-client?tab=readme-ov-file#installation

## Expected behavior

Successfully build and run the application

## Machine Information

 - OS: Fedora
 - Version: 40

alistair23:
The file does exist

```shell
$ ls -al /usr/bin/node-20
-rwxr-xr-x. 1 root root 28176 Sep 11 10:00 /usr/bin/node-20
```

matthewCmatt:
A few debugging ideas:
- Try upgrading to Node 22?
- Are you able to run `pnpm i`? `pnpm ui:dev`? `pnpm ui:lint`? `pnpm tauri`?
- Could you drop your versions from `node -v`? `pnpm -v`? `cargo version`?

matthewCmatt:
Closing as this is stale and can't reproduce.

Feel free to reopen if you still see this with latest node+pnpm+cargo versions

PR #493, opened 2025-03-07
Add devcontainer

ThatKalle:
Added an initial [development container](https://code.visualstudio.com/docs/devcontainers/containers) configuration.
_Tested on a Windows11 host, Docker Desktop, and a WSL2 Ubuntu installation._

It starts up, installs dependencies needed to run `pnpm run rust:dev`.
Also installs `meshtasticd` and `meshtastic CLI`.

With that it's possible to access the application, see a node.

I did try changing some text in the application and it works.

image

issue #471
issue #472


matthewCmatt:
Wow! This is great work. This was attempted in #491 but looks like that MR was closed.
Love the inclusion of meshtasticd, this will be hugely helpful for testing in CI/CD etc.

Could you add something to the readme that states the dev container only works on Linux? I spun it up on my Mac and ran into issues with USB forwarding, and would also probably run into problems with display output from GTK.

Could you also move the `config.yaml` to a directory other than the root? Maybe into `/.devcontainer/`? Or give it a more descriptive name?

ThatKalle:
> Could you also move the config.yaml to a directory other than the root? Maybe into /.devcontainer/? Or give it a more descriptive name?

Sorted! `meshtasticd` looks for it in the current directory, but it was possible to set a new location with `-c path` syntax.

> Could you add something to the readme that states the dev container only works on Linux? I spun it up on my Mac and ran into issues with USB forwarding, and would also probably run into problems with display output from GTK.

That is quite, interesting.
It's kind of supposed to work everywhere, and it doesn't do any USB forwarding.
I don't have access to Mac or Linux hardware to test on.

I myself run it on a Windows 11 host with Docker Desktop and WSL2 Ubuntu backend.

> This was attempted in https://github.com/meshtastic/network-management-client/pull/491

Oh neat! Didn't see that one.
I've copied most of the README content with some updates.

ThatKalle:
> display output from GTK

I have somewhere in the back of my head that playing with the `$DISPLAY` variable can help that flow.
I've added a RUN step in the Dockerfile, made no difference to me, still works both ways,, but that might very well be some Microsoft magic.

Also added a note at the top of the README, hopefully that can get updated as more devs give it a go 👍

matthewCmatt:
> I have somewhere in the back of my head that playing with the $DISPLAY variable can help that flow.
I've added a RUN step in the Dockerfile, made no difference to me, still works both ways,, but that might very well be some Microsoft magic.

Yeah WSL is pretty well-instrumented from my experience. Looks like [this thread](https://github.com/meshtastic/firmware/discussions/4278) has some similar issues to mine, so maybe related to some MacOS docker stuff being more locked down?

> Also added a note at the top of the README, hopefully that can get updated as more devs give it a go 👍

This note is perfect, thanks.

ThatKalle:
Updated to match https://github.com/meshtastic/firmware/discussions/4278, give it a go

matthewCmatt:
> Updated to match [meshtastic/firmware#4278](https://github.com/meshtastic/firmware/discussions/4278), give it a go

Still no dice. I'm getting the same SPI issue when trying to run the [standalone meshtasticd container](https://hub.docker.com/r/meshtastic/meshtasticd), so this may need some more involvement to debug.

In the meantime I'm spinning this up on an old Ubuntu machine to see if that works. Container build took about 40 minutes, and had to retry a few times to get the VSCode Server to install in the container.
I got a similar `Gtk-WARNING **: 15:15:22.896: cannot open display: :0` error

matthewCmatt:
> In the meantime I'm spinning this up on an old Ubuntu machine to see if that works. Container build took about 40 minutes, and had to retry a few times to get the VSCode Server to install in the container. I got a similar `Gtk-WARNING **: 15:15:22.896: cannot open display: :0` error

I ran the following on the Ubuntu host:
```
xhost +
```

This lets the display output through, but the initial container script fails. On finishing, I was able to run `pnpm rust:dev` and the app came through fine with the meshtasticd node showing up properly! If we add the above to the README, I think we can list Ubuntu as a tested environment

takov751:
I have just tested today on Fedora/Nobara 41 worked as it was meant to be, great work. I used a remote web connection to my node this time.

ThatKalle:
Hey @matthewCmatt , if you do find a moment.
I did read somewhere that one is supposed to use debian `bullseye` on apple silicon.
I don't know if you are running a apple silicon mac, but if so, maybe that change would solve the issues for you?

Adjust line 1 in .devcontainer/Dockerfile from `FROM mcr.microsoft.com/devcontainers/base:bookworm` to `FROM mcr.microsoft.com/devcontainers/base:bullseye` and try to Rebuild the container.

matthewCmatt:
@ThatKalle

Thanks for looking into this! I was able to get the devcontainer working fine by installing XQuartz for the display forwarding. This worked on Apple ARM fine without having to change the Debian version. I had to resize the window manually to get the virtual display to fill the whole window, but otherwise works well!

matthewCmatt:
Here's the guide I used to get it working https://gist.github.com/sorny/969fe55d85c9b0035b0109a31cbcb088

matthewCmatt (approved):

matthewCmatt:
@ThatKalle Thanks for implementing this!

PR #500, opened 2025-05-04
fix: disable Specta typescript gen and pin logging level

matthewCmatt:
This PR should fix the app's production build.

A couple issues addressed:
- Specta type generation is broken, so this disables it. This means the `src/bindings/index.ts` will not update on build. Previously this would be replaced with an empty file, which would then break the types used in the web app.
- Pins the logging level to "DEBUG". For some reason, in the production build was set to "TRACE" and this would make the app stall on startup.

Both of these issues should be addressed in another story.

This fixes local builds on my Mac using `pnpm tauri build` and our target `pnpm rust:build`. Not yet tested this on Linux builds

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -26,11 +26,7 @@ fn export_ts_types(file_path: &str) -> Result<(), TsExportError> {
     ts_with_cfg(file_path, &ts_export_config)
 }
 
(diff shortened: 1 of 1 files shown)


matthewCmatt:
Ran a build on Ubuntu 24.04 and the artifact ran fine

PR #515, opened 2025-07-06
Bluetooth Support

matthewCmatt:
Adds basic support for connecting to Meshtastic radios via Bluetooth LE.

Screenshot 2025-07-06 at 7 28 12 PM

- Uses the latest revision of `meshtastic/rust`. This will eventually be released, but currently it's just pinned to a commit hash.
- Includes an update to the Meshtastic protobufs in the client.

- [x] MacOS testing
- [ ] Linux testing
- [ ] Windows testing

### DIFF
--- a/src-tauri/src/ipc/commands/connections.rs
+++ b/src-tauri/src/ipc/commands/connections.rs
@@ -35,6 +35,16 @@ pub async fn request_autoconnect_port(
     Ok(autoconnect_port)
 }
 
+// Tauri command
+#[tauri::command]
+pub fn get_all_bluetooth() -> Result<Vec<String>, CommandError> {
+    debug!("Called get_all_bluetooth command");
+
+    let devices = vec!["mxie_2128".to_string(), "woaha".to_string(), "Freak out!".to_string()];
+
+    Ok(devices)
+}
+
 #[tauri::command]
 pub fn get_all_serial_ports() -> Result<Vec<String>, CommandError> {
     debug!("Called get_all_serial_ports command");
--- a/src-tauri/src/ipc/commands/connections.rs
+++ b/src-tauri/src/ipc/commands/connections.rs
@@ -7,6 +7,7 @@ use crate::packet_api::MeshPacketApi;
 use crate::state;
 use crate::state::DeviceKey;
 
+use btleplug::api::ScanFilter;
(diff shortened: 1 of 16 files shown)


src-tauri/src/ipc/commands/connections.rs
```diff
@@ -32,6 +34,58 @@ pub async fn request_autoconnect_port(
 
     Ok(autoconnect_port)
 }
+use btleplug::api::{Central, Manager as _, Peripheral as _};
+use btleplug::platform::Manager;
+use tokio::time;
+
+#[tauri::command]
+pub async fn get_all_bluetooth() -> Result<Vec<String>, CommandError> {
```
lukipuki, line 42: I will add more comments later, but this discovery function should go into the meshtastic Rust crate. I'll try to add it within next week.

You'd want only those devices that run the meshtastic service.

matthewCmatt: Definitely! This should be easy enough to pick up and drop in to the `rust` crate.

I'm not familiar with how to other clients detect if a bluetooth device "supports Meshtastic", but that would be great to add too.

It'd be nice if there was a similar functionality for discovering USB/serial devices too, which may even have a similar implementation.

src-tauri/src/ipc/commands/connections.rs
```diff
@@ -32,6 +34,58 @@ pub async fn request_autoconnect_port(
...
+    let adapters = manager.adapters().await
+        .map_err(|e| e.to_string())?;
+
+    let adapter = adapters
+        .into_iter()
+        .next()
+        .ok_or_else(|| "failed to find adapter")?;
+
+    // Start scanning for devices
+    adapter.start_scan(ScanFilter::default()).await
```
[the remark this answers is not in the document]
matthewCmatt: Beautiful, I dropped this here but this code will probably be replaced later once these are exposed in the `rust` lib.

matthewCmatt:
@lukipuki I cloned your repo and got it integrated, and it runs great on my MacOS!

I can check Linux and Windows later, but no complaints here!

lukipuki:
Thanks for checking! I'll merge soon then, as this is an improvement. We can fix bugs later.

PR #517, opened 2025-07-23
Add an "About" page with version info

matthewCmatt:
Adds an "About" page within the application settings that should help users report version information.

Screenshot 2025-07-23 at 9 06 14 AM


DaneEvans:
FW and any other details for the attached node is probably worth having on here too.

garthvh (approved):
Looks good

PR #519, opened 2025-07-23
Upgrade Biome to v2

matthewCmatt:
Most of this is just `pnpx @biomejs/biome check --write`

The important files are:
- .github/workflows/lint.yml
- biome.json
- package.json
- pnpm-lock.yaml


DaneEvans (approved):
Looks good

PR #525, opened 2025-09-03
fix(ui)!: :bug: demo to fix incorrect coordinates when creating waypoint

### LINKED ISSUE #487: Waypoints are added to the first location clicked
When adding a mesh waypoint, the user can right click anywhere on the map to open a popup that will let you create a waypoint.

If you right click to open the menu, and then right click again, elsewhere, a new popup will be created/moved to the new click location. As a user, I expect the waypoint created using this new popup to be at the new location. However, currently, the mesh waypoint will be added at the first location clicked.

In the below example video, the first click is in France and the second click is in Belarus. The waypoint ends up being added to France.

assanges:
As titled

Added key prop to `<CreateWaypointDialog />` for forced re-render/ remount. 
This demo should resolve the issue with incorrect coordinates (`lastRightClickLngLat`).

Fix #487

### DIFF
--- a/src/components/Map/MapView.tsx
+++ b/src/components/Map/MapView.tsx
@@ -269,7 +269,7 @@ export const MapView = () => {
         {(lastRightClickLngLat || editingWaypoint) && (
           <CreateWaypointDialog
             // use key prop to force re-render
-            key={`${lastRightClickLngLat?.lat}-${lastRightClickLngLat?.lng}-${Date.now()}`}
+            key={lastRightClickLngLat?.toString()}
             lngLat={lastRightClickLngLat ?? new LngLat(0, 0)}
             closeDialog={() => {
               setEditingWaypoint(null);


src/components/Map/MapView.tsx
```diff
@@ -268,6 +268,8 @@ export const MapView = () => {
 
         {(lastRightClickLngLat || editingWaypoint) && (
           <CreateWaypointDialog
+            // use key prop to force re-render
+            key={`${lastRightClickLngLat?.lat}-${lastRightClickLngLat?.lng}-${Date.now()}`}
```
assanges, line 272: Please replace the key as needed, cheers!

matthewCmatt, line 272:
```suggestion
            key={lastRightClickLngLat?.toString()}
```

matthewCmatt (commented):
Looks great, you hit the nail on the head!

matthewCmatt (approved):

PR #536, opened 2025-11-06
fix: resolve BLE device discovery serialization and Redux state issues

### LINKED ISSUE #534: Bluetooth connection on Windows
Bluetooth connnection to a radio on Windows is broken.

This may be a permission issue where Windows is blocking btleplug from discovering devices.

mdecourcy:
- Fix BluetoothConnectionCandidate deserialization in TypeScript The Rust backend returns BluetoothConnectionCandidate as a newtype struct, but TypeScript was casting it as plain strings. Added proper mapping to extract the string value from the struct.

- Add error handling for BLE scan failures Wrapped the getAvailableBluetoothDevices call in try-catch to handle BLE scanning errors gracefully and dispatch empty device list on failure.

- Add missing setAutoConnectBluetooth Redux action The action was being called in the codebase but wasn't defined in the device slice reducer. Added the reducer to properly manage autoconnect Bluetooth state.

Should resolve #534 but I don't have a PC so unable to validate on Windows. Validation passes on MacOS.

### DIFF
--- a/src/api/connection.ts
+++ b/src/api/connection.ts
@@ -25,9 +25,14 @@ export const requestAutoConnectPort = async () => {
 export const getAllBluetooth = async () => {
   const response = (await invoke("get_all_bluetooth", {
     request: {},
-  })) as { candidates: string[] };
+  })) as { candidates: Array<{ 0: string }> | string[] };
 
-  return response.candidates;
+  // Handle both formats: newtype struct { 0: string } and plain strings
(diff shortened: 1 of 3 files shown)


kairos17:
It doesn't work under Win10.  It can't even start do anything with BT. Just nothing is happening.
Windows 10 Pro
Ver.	22H2
OS set	19045.6456

PR #538, opened 2025-12-18
add russian translation

krotesk:
Add the remaining multilingual settings yourself.


immorrtalz:
Probably needs to be added to `src\i18n.ts` too - this way it works.

image

matthewCmatt:
Looks great, thank you! I'll check this out later in-app

